PR #474, opened 2019-05-05
[WIP] Use Gaussian instead of MvNormal

ChrisRackauckas:
Continuation of #221


andreasnoack:
Better to start over if we want to go this route.

PR #603, opened 2019-07-28
Add MCMCChains wrapper

Vaibhavdixit02:
@vjd the wrapper is there but there is some issue with the plotting which I haven't been able to diagnose. I might need to ask the Turing people regarding it.

@andreasnoack there is something weird going on in my local git history because of the last commit to master, I can't rebase, did you squash and merge or something?


Vaibhavdixit02:
Also see https://github.com/TuringLang/AdvancedHMC.jl/issues/89

vjd:
Does this come with the ability to parallelize the chains?

src/estimation/bayes.jl
```diff
@@ -132,17 +132,36 @@ struct BayesMCMC <: LikelihoodApproximation end
...
   dldθ(θ) = logdensitygrad(bayes, θ)
   l(θ) = logdensity(bayes, θ)
-  metric = DiagEuclideanMetric(length(θ_init))
+  metric = DiagEuclideanMetric(length(vparam_aug))
   h = Hamiltonian(metric, l, dldθ)
-  prop = AdvancedHMC.NUTS(Leapfrog(find_good_eps(h, θ_init)))
-  adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, prop.integrator.ϵ))
-  samples, stats = sample(h, prop, θ_init, n_samples, adaptor, n_adapts; progress=true)
-  BayesMCMCResults(bayes, samples, stats)
+  prop = AdvancedHMC.NUTS(Leapfrog(find_good_eps(h, vparam_aug)))
```
ChrisRackauckas: we should make these kinds of things options in the future

ChrisRackauckas:
Not exactly, but it's pretty clear how to parallelize the chains (just parallelize the call to produce the chains), and IIRC MCMCChains.jl has a constructor from multiple chains so the analysis is easier.

PR #606, opened 2019-07-31
Fix README example

ChrisRackauckas:


vjd:
do we need to change the results to reflect the `Random.seed()` results?

ChrisRackauckas:
single_sub

pop_sim

vpc

highdose

PR #624, opened 2019-08-07
Fix depwarns in nca test files.

pkofod:
This fixes depwarns from DataFrame changes in the NCA tests.


pkofod:
When fixing one things breaks another :) I'm debugging...

pkofod:
Okay, so in the middle of fixing pumas, this was merged and tagged in a minor release https://github.com/JuliaData/DataFrames.jl/pull/1907 . I'll have to fix the errors it introduces for us.

PR #625, opened 2019-08-07
Standardise fit signature for ML and Bayes, fixes #620

### LINKED ISSUE #620: Make initial value passing consistent between ML and Bayesian fit methods
We currently pass the initial values as the third positional argument to the `fit` methods for ML while it's an optional fourth argument in the Bayesian `fit` method. It should be made consistent. The simpler solution is to deprecate the current Bayesian `fit` signature and introduce a new one similar to the ML `fit` methods.

This is somewhat related to #424 since the current Bayesian `fit` method relies on initial values from the domains.

Vaibhavdixit02:


src/estimation/bayes.jl
```diff
@@ -130,7 +130,7 @@ struct BayesMCMC <: LikelihoodApproximation end
 #   BayesMCMCResults(bayes, chain, tuned)
 # end
 
-function Distributions.fit(model::PumasModel, data::Population, ::BayesMCMC, θ_init=init_param(model),
+function fit(model::PumasModel, data::Population, ::BayesMCMC, θ_init=init_param(model),
```
ChrisRackauckas: Distributions.fit(model,data,param::BayesMCMC ...)

andreasnoack:
Would be nice with a deprecation of the old signature.

src/estimation/bayes.jl
```diff
@@ -157,6 +157,11 @@ function Distributions.fit(model::PumasModel, data::Population, ::BayesMCMC, θ_
   BayesMCMCResults(bayes, samples_, stats)
 end
 
+function Distributions.fit(model::PumasModel, data::Population, param::NamedTuple, ::BayesMCMC,
```
ChrisRackauckas: it would be nice to include a dispatch without `param`

andreasnoack: Using `init_param`?

Vaibhavdixit02: @ChrisRackauckas if you look at the arguments of `fit` defined in Line 133, it allows calling `fit` without passing the parameters. So, maybe we shouldn't deprecate that?

ChrisRackauckas: yeah that would work.

PR #626, opened 2019-08-08
Use map instead of broadcast in read_pumas

asinghvi17:
Workaround for JuliaData/DataFrames.jl#1907

cc @pkofod @ChrisRackauckas - will this patch produce the correct behaviour?


asinghvi17:
Doesn't look like it, tested with an example from the tutorials.

asinghvi17:
Gitlab isn't running - is the CI active?

ChrisRackauckas:
It should be. Give it a sec.

pkofod:
does this fix the error in #624 ?

asinghvi17:
Think so :) read_pumas works with latest DF at least

andreasnoack:
It looks like this isn't right. The NLME tests are completely off

ChrisRackauckas:
I don't immediately see why Refing those last ones wouldn't work, but if the tests fail then it needs to get looked into further.

andreasnoack:
I think the implementation here assumes broadcasting behavior but `map` doesn't broadcast. Maybe simpler to use a comprehension.

asinghvi17:
Will make the change.  Has Gitlab still not triggered?

andreasnoack:
GitLab did run the tests. That was how I realized that there was a problem with the implementation. For some reason, the webhook hasn't gone through which is why the test status is still pending here. You can (hopefully) see the test results here https://gitlab.com/PumasAI/Pumas-jl/pipelines/75347608

asinghvi17:
Superseded by #628, thanks @andreasnoack!

PR #627, opened 2019-08-08
Implement convergence type recipe for FittedPumasModel

asinghvi17:
Implements a Plots type recipe, which plots the convergence.  This can be changed to a user recipe `convergence` or something like that.

### DIFF
--- a/src/estimation/diagnostics.jl
+++ b/src/estimation/diagnostics.jl
@@ -481,3 +481,45 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
 
   df = hcat(pred_df, res_df, ebes_df)
 end
+
+@recipe function f(f::FittedPumasModel)
+  # extract the parameter space vectors, and allow them to be plotted as timeseries.
+
+  trf  = totransform(f.model.param)         # get the transform which has been applied to the params
+  itrf = toidentitytransform(f.model.param) # invert the param transform
+
+  paramtrace = hcat(TransformVariables.inverse.(Ref(itrf), TransformVariables.transform.(Ref(trf), getindex.(getproperty.(f.optim.trace, :metadata), "x")))...)
+  paramnames = [] # empty array, will fill later
+
+  legend := :none # no legends - they clutter the plot up too much.
+
+  for (paramname, paramval) in pairs(f.param) # iterate through the parameters
+    # decompose all parameters (matrices, etc.) into scalars and name them appropriately
+    _push_varinfo!(paramnames, [], nothing, nothing, paramname, paramval, nothing, nothing)
+  end
+
+  # @show paramtrace paramnames
+  layout --> length(paramnames) + 1
+
+  @series begin
+   title := "Observed value"
+   subplot := length(paramnames) + 1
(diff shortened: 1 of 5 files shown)


asinghvi17:
I've changed the plotting API to reflect that plotting in Pumas will need to be extensible.  Therefore, my proposal for the internal API is this:
- Each "type" will implement methods to access plottable data which are required for the various plot types.  This will allow the creation of an extensible plotting framework, which can be used by any type which has the required methods implemented for it.
- These functions will be documented for each plot type, so that using them later becomes easy. 
- Each plot type (convergence, qq, etc. ) will then be able to use "multiple dispatch" to get plottable data easily.  This will simplify recipes and decrease the number of changes that need to be made.

We also need to look at how we can avoid conflicts with StatsPlots if possible (no overlapping recipes, et cetera).

asinghvi17:
Apologies for the dual commits, should this be squashed before merging?

pkofod:
> Apologies for the dual commits, should this be squashed before merging?

Don't worry about that.

src/estimation/diagnostics.jl
```diff
@@ -481,3 +481,106 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
...
+################################################################################
+#                              Plotting functions                              #
+################################################################################
+
+########################################
+#   Convergence plot infrastructure    #
+########################################
+
+"""
+    _observedvalues(obj)
```
vjd: This should be `objectivefunctionvalues`

src/estimation/diagnostics.jl
```diff
@@ -481,3 +481,106 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
...
+
+Returns the observed function values during optimization.
```
vjd: objective function

src/estimation/diagnostics.jl
```diff
@@ -481,3 +481,106 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
...
+Must return a `Vector{Number}`.
+"""
+_observedvalues(f::FittedPumasModel) = getproperty.(f.optim.trace, :value)
```
vjd: _objectivefunctionvalues

src/estimation/diagnostics.jl
```diff
@@ -481,3 +481,106 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
...
+
+    data  = _convergencedata(arg) # again, a wrapper function.
+
+    layout --> good_layout(length(names) + 1)  # good_layout simply overrides the Plots layouter for n <= 4
+
+    legend := :none # clutters up the plot too much
+
+    # also plot the observed function value from Optim
+    @series begin
+        title := "Observed value"
```
vjd: Objective Function Value

asinghvi17: Will go through all of these and change, thanks!

asinghvi17:
Not sure whether that would be the right approach if we were to make each folder a submodule?

PR #628, opened 2019-08-08
Use comprehension instead of broadcasting in read_pumas

andreasnoack:
Since master is failing we better fix this right away.

Supersedes #626.


andreasnoack:
Tests pass. I'll merge right away to make master green again.

PR #629, opened 2019-08-09
[RFC] "Reparameterize" random effects internally

### LINKED ISSUE #570: Use TransformVariables when extracting η
So far, we have been hard coding the name of random effects to `η` in the estimation code. Eventually, we should use the `TransformVariables` mechanism for this. A very nice consequence of fixing this issue would be that we could introduce the convention that separate random effects are independent and write the more common diagonal as e.g.
```julia
ηCL ~ Normal(...)
ηV ~ Normal(...)
```
However, all the code for computing the Laplace approximation refers to the joint covariance matrix of all the random effects so we'd have to introduce functionality to compute this. Shouldn't be too hard, though.

### LINKED ISSUE #599: Transform random effect in likelihood calculation
In the computation of the marginal likelihood, we have terms of the form
image
which fails when `Ω` becomes singular. @dmbates has pointed out that we can rewrite the terms to
image
This avoids solving with `Ω` and, therefore, it isn't numerically a problem when `Ω` becomes singular. Hence, it might improve estimation if we change the likelihood calculation to use the scaled random effects.

andreasnoack:
This implements the suggestion in #570. However, I realized that using orthogonal random effects is actually even nicer than expected. It happens that if we let the vector representation of the random effect `vrandeffs` (now renamed to `vrandeffsorth`) represent the orthogonalized random effect such that `randeffs = (η=sqrt(Ω)*vrandeffsorth,)` then all occurrences of `Ω` disappear from the mathematical expressions. As an example of this, the laplace approximation of the marginal likelihood used to be computed as
```julia
logdet(Ω) + logdet(inv(Ω) + W) + η'*(Ω\η)
```
is now simply
```julia
logdet(I + W) + vrandeffsorth'vrandeffsorth
```
because `W` is now the Hessian with respect to `vrandeffsorth`.

Furthermore, we only solve with `Ω` when starting out from `η` so we can completely avoid solving with `Ω` or compiting the inverse during estimation. Hence, it's not a problem if `Ω` becomes singular. It should make estimation more robust.

Finally, with these changes we use `TransformVariables` systematically for converting between `randeffs` and `vrandeffsorth` so, as a side effect, it is now possible to specify multiple `η`s. It is then assumed that they are independent. Hence, I could change the last theophylline teset to
```julia
@param begin
    ...
    ω²Ka ∈ RealDomain(lower=0.0)
    ω²CL ∈ RealDomain(lower=0.0)
    ...
end

@random begin
    ηKa ~ Normal(0.0, sqrt(ω²Ka))
    ηCL ~ Normal(0.0, sqrt(ω²CL))
end
```

This is mostly complete and things work. What is missing the the EBE uncertainty calculation. I'll have to think a little more about that one. Then it also appears that `ηshrinkage` isn't invariant under this transformation which bothers me. All the other diagonstics were invariant such that I simply had to delete `Ω` from the expression but `ηshrinkage` is different. @jony5525 where does `η`-shrinkage come from and do you know why the inventors didn't scale the complete empirical covariance matrix with the complete `Ω` but only instead only used the diagonals?

The way I'm currently using `totransform` to define transformations for `MvNormal` and `Normal` might also clash with the way we use `totransform` to allow using the ML fitting method for models with priors specified. I'd also have to think a little bit more about that. Maybe we want to use a separate function for the transformation of random effects.

Fixes #570
Fixes #599


jony5525:
> logdet(I + W) + vrandeffsorth'vrandeffsorth

That's great! Did you see any speed improvements in the actual likelihood calculations or do you see most of the improvements in the stability of the population parameter estimation?

Eta-shrinkage comes from e.g. 
[Savic_Karlsson](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2758126/pdf/12248_2009_Article_9133.pdf) but I'm not sure that it make sense to scale the shrinkage by the correlation/covariances.

pkofod:
> Finally, with these changes we use `TransformVariables` systematically for converting between `randeffs` and `vrandeffsorth` so, as a side effect, it is now possible to specify multiple `η`s

Also no random effects?

andreasnoack:
@jony5525 I haven't yet compared timings. Will share them once I've run them. Thanks for the reference. Will take a look. It's a bit of a shame if there is no way around constructing `Ω` since it would probably be the only computation in Pumas for which it was needed.

@pkofod Not it the current form and I'm not sure it will be desirable. The failure is because we try to differentiate wrt zero-length vectors. It's probably better simply to avoid calling the marginal likelihood functions.

andreasnoack:
I found a way to handle η-shrinkage so now the only remaining item is the uncertainty of the random effects.

pkofod:
(I really like the individual etas here instead of the vectors tbh)

andreasnoack:
I found a way to handle the EBE uncertainty, so I think this one is ready now. I did a few benchmarks of some the models in theophylline.jl and bolus.jl and it looks like this change also has a speed advantage.

#### Run 4. FOCEI (closed form) from theophylline.jl
```julia
# Old:
julia> @btime fit(theopmodel_focei, theopp, param, Pumas.FOCEI());
  114.534 ms (1334028 allocations: 108.45 MiB)

# New:
julia> @btime fit(theopmodel_focei, theopp, param, Pumas.FOCEI());
  88.954 ms (1243978 allocations: 107.44 MiB)
```

#### Laplace estimation of the proportional error model from bolus.jl
```julia
Old:
julia> @btime fit(mdl_proportional[dyntype], data, param_proportional, Pumas.Laplace())
  17.969 s (111363494 allocations: 16.99 GiB)

julia> @btime fit(mdl_proportional[dyntype], data, param_proportional, Pumas.Laplace());
  14.034 s (99919939 allocations: 15.58 GiB)
```

pkofod:
LGTM. I've merged master into this, and will merge if it completes the test suite.

pkofod:
🍾

dmbates:
Great work on that, @andreasnoack That was the point I was trying to make earlier.

PR #634, opened 2019-08-12
Implement plotting utilities

asinghvi17:
- [x] Remove the dependency on Loess.jl, in favour of using DataInterpolations.jl instead.
- [x] Alias the type recipe for FittedPumasModel to a userplot called `convergence`
- [x] Implement a new user plot `etacov`
- [x] Implement a new user plot `resplot`


asinghvi17:
With the README example's FittedPumasModel:
`etacov(res)` results in:
plt1

I'll add a docstring once the API has been fleshed out.

cc @vjd

asinghvi17:
With StatsPlots as a dependency, Pumas takes 153 seconds to precompile + load, and 60 seconds to load; compared to 80 and 40 earlier, this is a serious regression.

We could perhaps place the recipes which require StatsPlots (probably most of the complex ones) into a `@requires` block?

vjd:
> With the README example's FittedPumasModel:
> `etacov(res)` results in:
> plt1
> 
> I'll add a docstring once the API has been fleshed out.
> 
> cc @vjd

We need a way to allow the user to specify if the covariate is continuous or categorical. in the above example, `isPM` should be categorical in nature.

vjd:
> precompile + load, and 60 s

@ChrisRackauckas @andreasnoack @pkofod - I seriously think that we should have a separate package for `PumasPlotting` rather than shoving everything in here. What do you think?

ChrisRackauckas:
This PR doesn't have the plotting in here.

asinghvi17:
It does, in the file `plotting/stat_plots.jl` - it's just locked behind Requires, so using StatsPlots should "unlock" those recipes for you.

vjd:
so, are we going with keeping things in here and requiring the user to load StatsPlots?

asinghvi17:
Having a separate package would in my view be the ideal solution, but for now the user will have to load StatsPlots to get access to any of the "advanced" plotting recipes.

asinghvi17:
> We need a way to allow the user to specify if the covariate is continuous or categorical. in the above example, isPM should be categorical in nature.

It is - that's why it's a boxplot - but I've introduced a kwarg `catmap`, which will allow the user to provide a NamedTuple of `colname = true` (for categorical) or `false` (for continuous).  In the future, it would also be nice to have a keyword to control plot types per column, but that's something that can be done later.

As for the eta - I just set the ylabel to `"η"`.

andreasnoack:
> so, are we going with keeping things in here and requiring the user to load StatsPlots?

I think having a `PumasPlots` package makes sense and should be fairly easy to communicate to ordinary users.

> We could perhaps place the recipes which require `StatsPlots` (probably most of the complex ones) into a @requires block?

I'd prefer not to. Requires doesn't work well with precompilation and makes it impossible to wrap the package in a stand-alone app.

> We need a way to allow the user to specify if the covariate is continuous or categorical.

@vjd Couldn't we use https://github.com/JuliaData/CategoricalArrays.jl for this. It's the standard way of specifying categorical variables in Julia's StatsModels/GLM

asinghvi17:
My fear is that the user may not wrap the input in a categorical array, but pass in something like a String, which we'd need to convert internally.  That's why I implemented this heuristic here
https://github.com/PumasAI/Pumas.jl/blob/af3b46ef6082595af5bea67c833cdfa15a01446f/src/plotting/plotting.jl#L5-L10
so that, while the user has the option of specifying the categorical array, they can allow the heuristic to decide that as well.

andreasnoack:
I'm not sure what the better solution is here but we should keep in mind that this problem is in no way specific to Pumas. It's shared among all stats and plotting packages in Julia so we should think about how we handle this in that context.

I think the first four definitions are okay but I don't like the fifth. It's too arbitrary and often we avoid changing behavior based on *values*. But again, we should think about this in relation to the larger ecosystem while, of course, insist that it works well for our application.

asinghvi17:
What remains in this PR is grouping plots by "scenario".  IMO, we should provide a `Group(:isPM)` constructor a la StatsMakie, which would then have an API like `resplot(res, Group(:isPM))` which would create two panels.  I'm going to ask a couple of questions to the Plots maintainers about how I can do the layouting, but if that ends up not being possible, then I will probably end up hacking together something for the time being - it may look a bit ugly, though.

asinghvi17:
I've implemented grouping in a sort of roundabout way, by putting the onus on the user.  Now, the _user_ has to group `DataFrame(inspect(res))` using DataFrames' functions.  Then, the subdataframes (**not** the grouped dataframe) can be passed to the plotting functions.

Since advanced layouts aren't trivial to create in Plots, I'll create a function with signature `groupedplot(plotting_func, dataframe, cols; sort=false, skipmissing=false, titles = [], plot_kwargs...)`.  The input dataframe will be grouped by `cols`, and then each element of it will be passed to titles.  Then, the plotting function will be called for each subdataframe.

asinghvi17:
Test failure is because this branch is outdated.  Will try a rebase; if not, then I'll open a new branch with this patch.

asinghvi17:
I think this PR is in a good state now - most of the meaty changes have been moved to PumasPlots, so this really just moves the type recipe for FittedPumasModel into a userplot (in case we want to plot something else with it), and adds some utility functions.

@vjd I can add the readme examples in this PR now, or later once PumasPlots is registered.  Its grouping functionality is almost done, just need to figure out a good API.  I've messaged you on Slack about that.

asinghvi17:
With this PR, and PumasPlots master, here's a summary of new plot types.

- `convergence(res::FittedPumasModel)` - plots the optimization trajectory of all variables
- `etacov(res::FittedPumasModel)` - plots the covariates of the model against the etas.  Keyword arguments are described in the docstring - essentially, you can use any column in `DataFrame(inspect(res))` to plot.  
- `resplot(res::FittedPumasModel)` - plots the covariates of the model against their residuals, determined by the approximation type.  Accepts many of the same kwargs as `etacov`.
- `corrplot(empirical_bayes(res))` - overload of the `StatsPlots` corrplot, for etas.

PR #636, opened 2019-08-12
Insert a dummy/catch-all method for fitting a single subject with a m…

pkofod:
…ethod for population estimation. There are no etas to identify off of a single subject, so just never allow this to be attempted.

Methods more specific than this one will be written for the code that does single subject estimation later.

Right now the error is
```
ERROR: MethodError: no method matching fit(::PumasModel{ParamSet{NamedTuple{(:θ, :Ω, :Σ),Tuple{VectorDomain{Array{Float64,1},Array{Float64,1}},ConstDomain{Diagonal{Float64,Array{Float64,1}}},ConstDomain{Float64}}}},getfield(Main, Symbol("##3#8")),getfield(Main, Symbol("##4#9")),getfield(Main, Symbol("##5#10")),ImmediateAbsorptionModel,getfield(Main, Symbol("##6#11")),getfield(Main, Symbol("##7#12"))}, ::Subject{NamedTuple{(:dv,),Tuple{Array{Union{Missing, Float64},1}}},Nothing,Array{Pumas.Event{Float64,Float64,Float64,Float64,Float64,Float64,Int64},1},Array{Float64,1}}, ::NamedTuple{(:θ, :Ω, :Σ),Tuple{Array{Float64,1},Diagonal{Float64,Array{Float64,1}},Float64}}, ::Pumas.FOI)
Closest candidates are:
  fit(::Type{StatsBase.Histogram}, ::Any...; kwargs...) at /home/pkofod/.julia/packages/StatsBase/DyWPR/src/hist.jl:319
  fit(::StatsBase.StatisticalModel, ::Any...) at /home/pkofod/.julia/packages/StatsBase/DyWPR/src/statmodels.jl:152
  fit(::Type{D<:Distribution}, ::Any...) where D<:Distribution at /home/pkofod/.julia/packages/Distributions/tfkz4/src/genericfit.jl:34
  ...
Stacktrace:
 [1] top-level scope at none:0
```
which is not nice.


andreasnoack:
Couldn't we just leave out the `approx` argument for the `fit(,,::Subject,...)` method?

pkofod:
> Couldn't we just leave out the `approx` argument for the `fit(,,::Subject,...)` method?

Yes, you're right - I actually just came to the same conclusion. However, it's still nice to explicitly catch the rror which is what this PR specifically tries to do. When we fix up the single subject estimation, we want to point the users there in the error message introduced here.

ChrisRackauckas:
I like the explicit error catching, but edit the error message here to mention that fitting with a single subject shouldn't have a likelihood approximation (even though we don't have it implemented yet, I think there's a 70% chance we'd forget to update the error message)

pkofod:
> I like the explicit error catching, but edit the error message here to mention that fitting with a single subject shouldn't have a likelihood approximation (even though we don't have it implemented yet, I think there's a 70% chance we'd forget to update the error message)

I see your point. I'll change it, but Andreas and I are currently working on this :)

pkofod:
Good, so now it errors on likelihoodapproximations, and (for now) errors on the fit call without a LikelihoodApproximation as well.

PR #637, opened 2019-08-13
Naive estimation methods

### LINKED ISSUE #510: fit single subject data
using any of the theophylline tests, can we fit data only for a single subject? So, no mixed effects, only the population parameters that define the analytical solution and perhaps the error model but no `omegas`?

`single = fit(theopmodel_focei, theopp[1], param, PuMaS.FOCEI())`

what needs to be done to make this work?
1. fix `omega` to zero?
2. remove `omega` and `@random` from the model and hence all the `etas` in the `@pre`?

pkofod:
fixes #510 

The `NoApprox` struct is just because the third element of `FittedPumasModel` has to be a `LikelihoodApproximation`.

Just to show an example of a histogram of estimates over individual fits, look at `tvcl` (data generating parameter is 4.0)
tvcl

I may want to test something in the test set, suggestions appreciated!


test/nlme/single_subject.jl
```diff
@@ -36,7 +31,54 @@ param = init_param(mdsl1)
...
+
+simdf = DataFrame(obs)
+simdf.cmt .= 1
+
+data = read_pumas(simdf, time=:time,cvs=[:isPM, :wt])
+
+res = fit(model,data[1],param)
+trf = Pumas.totransform(model.param)
+fits = [fit(model, dat, param) for dat in data]
+# test something?
```
vjd: I will run this dataset in other software and let you know the stats around the individual fits

test/nlme/single_subject.jl
```diff
@@ -36,7 +31,54 @@ param = init_param(mdsl1)
...
+obs = simobs(model, pop_with_covariates, param, obstimes=0:1:120)
+
+simdf = DataFrame(obs)
+simdf.cmt .= 1
+
+data = read_pumas(simdf, time=:time,cvs=[:isPM, :wt])
+
+res = fit(model,data[1],param)
+trf = Pumas.totransform(model.param)
+fits = [fit(model, dat, param) for dat in data]
```
vjd: The fits are super fast! Wow.
Do we have a way to post-process the results? As in means of parameters across the subjects.

pkofod: No, but I will write those tools. I also need to make the other tools work I’d they don’t (infer etc).

They probably fit quickly because they circumvent whole machinery for random effects. This should only be in compilation though .

pkofod:
Good, so now this works and re-uses the existing infrastructure for fitting. 

There is a caveat, but it's something we'll have to accept for now. You have to *specify* the model without the random effects. This unfortunately also means some compilation cost compared to a hypothetical case where we could "ignore etas" if the same script works with the model with and without the etas. The biggest problem right now is that there is nothing in the code/dsl that allows us to tag Omega as the variance of the ebes. This means that the optimizer will try to chose optimal omega entries . We could easily zero out the ebes themselves, but that doesn't stop the optimizer from chosing Omega-entries.

vjd:
>This pr could use a show method for the vector of fitted models btw :) You'll see why if you try to print it.

As discussed on slack, for the individual fits or naive methods:
1. `DataFrame([fit.param])` should convert the named tuple of optimized parameters into a data frame
2. For `NaivePooled()` and `NaiveAveraged()` we should get a dataframe of 1 row and for `TwoStage()` we should get as many rows as number of subjects.
3. For the first two methods, we just transpose the dataframe and print it out using
```
julia> stackdf(DataFrame([resparam]),[:tvcl,:tvv,:Ω, :σ_prop])
4×2 DataFrame
│ Row │ variable │ value                │
│     │ Symbol   │ Any                  │
├─────┼──────────┼──────────────────────┤
│ 1   │ tvcl     │ 11.5                 │
│ 2   │ tvv      │ 50                   │
│ 3   │ Ω        │ [0.04 0.0; 0.0 0.04] │
│ 4   │ σ_prop   │ 0.001                │
```
4. For the `TwoStage()`, before stacking the data frame, we can compute column statistics such as mean and standard deviation and then do the transpose.

pkofod:
Did you get to have a look vijay?

src/Pumas.jl
```diff
@@ -61,4 +62,5 @@ export @model, @nca, @tvcov
 export fit, stderror, vcov, aic, bic, deviance, informationmatrix
 export infer, inspect
 export vpc, vpc_obs
+export mean, std, var
```
ChrisRackauckas: what's the reasoning for this?

src/estimation/likelihoods.jl
```diff
@@ -4,6 +4,8 @@ const DEFAULT_RELTOL=1e-6
 const DEFAULT_ABSTOL=1e-12
 
 abstract type LikelihoodApproximation end
+struct NaivePooled <: LikelihoodApproximation end
+struct TwoStage <: LikelihoodApproximation end
```
ChrisRackauckas: @vjd is this verbiage people would know it by? Because in parameter estimation of ODEs two-stage usually stands for a different method.

src/Pumas.jl
vjd: When we are doing an individual fitting of each subject in a dataset, then we end up getting an array of parameters. Taking basic stats on these functions will help us generate the parameter vector that can be used as initial estimates for the more complex algorithms

src/estimation/likelihoods.jl
vjd: Yes, this is common in pharmacometrics. TwoStage involves, fitting each subject individually and then summarizing the stats (mean and variability) in that population across all subjects

src/Pumas.jl
ChrisRackauckas: Why not calculate this for the user as the suggested fits then? That would make the API here match the other methods.

ChrisRackauckas:
We weren't done discussing the API here... I guess @pkofod  can do a follow up PR.

andreasnoack:
Sorry, didn't see your last comment had arrived. Let's discuss here and do a followup PR, if needed.

ChrisRackauckas:
Yeah it's not a big deal, this was good enough to merge. Just after that last comment from @vjd , it seemed like we might as well take advantage of that and make this match the normal fitting interface. It can give back all of the data, but compute the means and variances and print like a `FittedPumasModel` to act more uniformly with the others. If that's the reason for the mean/std exports, we might as well just compute that and make the nice printout because it's cheap to do anyways.

PR #640, opened 2019-08-14
IVIVC API changes

sipah00:


vjd:
@sipah00 can you please add some tests to this when you can! Its time to get this wrapped up!

sipah00:
@vjd I have updated the tests. @ChrisRackauckas please have a look at function `do_ivivc` and see if there can be any improvement or optimization.

src/ivivc/IVIVC.jl
```diff
@@ -24,13 +24,13 @@ include("utils.jl")
 include("model_validation.jl")
 include("main.jl")
 
-export VitroSubject, VitroPopulation
-export VivoSubject, VivoPopulation
-export read_vitro, read_vivo
+export VitroForm, VitroData
```
ChrisRackauckas: InVitro InVivo

src/ivivc/IVIVC.jl
```diff
@@ -24,13 +24,13 @@ include("utils.jl")
...
+export VivoForm, VivoData
+export read_vitro, read_vivo, read_uir
```
ChrisRackauckas: uir?

src/ivivc/data_parsing.jl
```diff
@@ -55,18 +55,45 @@ function ___read_vivo(df; id=:id, time=:time, conc=:conc, form=:form, dose=:dose
...
+  pop = VivoData(identity.(ncas))
   return pop
 end
 
 
+
+# Uir Data
+read_uir(file::AbstractString; kwargs...) = read_uir(CSV.read(file); kwargs...)
+
+function read_uir(df, kwargs...)
```
ChrisRackauckas: this is unnecessary

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
-function do_ivivc(vitro_batch, ref_vivo_pop, vivo_pop;
-                    vitro_m=nothing,
+function do_ivivc(vitro_data, uir_data, vivo_data;
+                    vitro_model=nothing,
                     vitro_model_metric=:aic,
-                    ref_vivo_model=nothing,
+                    uir_frac = 1.0,
                     deconvo_method=:wn,
-                    corr_m=:two)
+                    ivivc_model=:two)
```
ChrisRackauckas: we will want to expand this later, but it's fine for now.

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
   # model the vitro data
-  # TODO: if vitro model is not specified by user then try all available models
-  # and select the best one on the basis of aic, aicc and bic
-  if vitro_m == nothing
-    for (form, prof) in vitro_batch[1]
-      try_all_vitro_model(prof, metric=vitro_model_metric)
-    end
+  if vitro_model === nothing
+    error("Not implemented!!")
```
ChrisRackauckas: make it required then. This is just messy.

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
-  ka, kel, V = 0.0, 0.0, 0.0
-  if ref_vivo_model == nothing
-    ref_vivo_model = lowercase(first(keys(ref_vivo_pop[1]))) == "solution" ? (:bateman) : (:iv)
-  end
-  
-  for (form, prof) in ref_vivo_pop[1]
-    vivo_model(prof, ref_vivo_model)
-    ka, kel, V = prof.pmin
-  end
+  uir_model = lowercase(uir_data.form) == "solution" ? (:bateman) : (:iv)
```
ChrisRackauckas: This seems a little odd to have embedded in the data format, while the other model choices are function arguments

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
+  ka, kel, V = uir_data.pmin
 
   # get the absorption profile from vivo data
-  # TODO: add if-else for all kind of methods (model dependent, model independent and numerical deconvolution)
-  # if none of the available methods are provided then we can run all methods and select the one which works
-  # best with dissolution data (in vitro data) based on validation
-
-  abs_vect = Vector{Dict{Any, Vector}}(undef, length(vivo_pop))
-  for i = 1:length(vivo_pop)
+  all_fabs = Vector{Dict{Any, Vector}}(undef, length(vivo_data))
```
ChrisRackauckas: This shouldn't need any dictionaries? Also, this could explode compile times. Let's do this right.

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
-    if corr_m == :two
-      corr_model(x, p) = @. p[1] * vitro_batch[1][form](x * p[2])
-      p1, p2 = Curvefit(fabs, vivo_pop[1][form].time, corr_model, 
-                              [0.8, 0.6], LBFGS()).pmin
-      abs_scale += p1; t_scale += p2
-      c += 1
+  avg_fabs = _avg_fabs(all_fabs)
+  # optimization
+  if ivivc_model == :two
+    m = (form, time, x) -> x[1] * vitro_data[1][form](time * x[2])
```
ChrisRackauckas: yeah, even with this PR... can we just have some prebuilt functions, but allow users to pass in any `(x,t)` function?

Also, this shouldn't need to take in `form`, just `data`. That can be abstracted for the user.

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
+  end
+  function errfun(x)
+    err = 0.0
+    for (form, prof) in vivo_data[1]
+      err = err + mse(m(form, prof.time, x), avg_fabs[form])
     end
+    return err
+  end
+  # intial estimates of params and bounds
+  p = [0.8, 0.5, 0.6, 0.6]
```
ChrisRackauckas: defaults, but not baked in...

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
+  opt_alg = LBFGS()
+  od = OnceDifferentiable(p->errfun(p), p, autodiff=:finite)
+  mfit = Optim.optimize(od, lb, ub, p, Fminbox(opt_alg))
+  pmin = Optim.minimizer(mfit)
+  do_ivivc{typeof(ka), typeof(pmin), typeof(all_fabs), typeof(all_auc_inf)}(vitro_data, uir_data, vivo_data, ka, kel, V, vitro_model, deconvo_method, all_fabs, all_auc_inf,
+                          m, opt_alg, p, ub, lb, pmin)
+end
+
+# main function for prediction by estimated IVIVC model
+function prediction(A::do_ivivc, form)
```
ChrisRackauckas: too generic of a name

src/ivivc/main.jl
```diff
@@ -1,92 +1,119 @@
...
 # do_ivivc(; vitro_data::AbstractString, ref_vivo_data::AbstractString,
 #   vivo_data::AbstractString, kwargs...) = do_ivivc(read_vitro(vitro_data, kwargs...),
 #                       read_vivo(ref_vivo_data, kwargs...), read_vivo(vivo_data, kwargs...),
 #                       kwargs...)
 
-struct do_ivivc{pType}
-  vitro_pop::VitroPopulation
-  ref_vivo_pop::VivoPopulation
-  vivo_pop::VivoPopulation
+struct do_ivivc{pType, paramsType, fabsType, aucType}
```
ChrisRackauckas: IVIVCModel

src/ivivc/main.jl
sipah00: IVIVC model involves fitted function of Vitro data, it is not just any (x, t), if we want to make it any `(x, t)` then we have to treat `t` as function to fitted vitro model (inside our function we have to do something like this `t -> Fdiss(t)`) 
a example of IVIVC Model => FAbs(t) = c1 * Fdiss(t * c2 - c3) + c4, where Fdiss function we will get after vitro modeling.

src/ivivc/main.jl
sipah00: 👍

src/ivivc/main.jl
sipah00: any suggestion

src/ivivc/main.jl
sipah00: yeah, we need something better for initial estimates of params

src/ivivc/IVIVC.jl
sipah00: :+1:

src/ivivc/data_parsing.jl
sipah00: I added this function to do some cleaning or other things (`dropmissings`) before passing it to `__read_uir`

sipah00: thought to add these things later

src/ivivc/main.jl
sipah00: 👍

src/ivivc/main.jl
sipah00: 👍

src/ivivc/main.jl
sipah00: this is just to get the formulation type of `uir data`, if it is solution type then we have to fit `bateman` and if it is `iv` type then we have to fit exponential decaying model, I'm not sure if this can be other than these two types then we have to switch to general function.

src/ivivc/main.jl
sipah00: can you suggest how can we map formulation to data except dict, may be some clever use of `Enum` (I'm not sure), I did this so that user can easily get the data for a formulation. `all_fabs[idx]["formulation_type"]`.

src/ivivc/plot_rec.jl
```diff
@@ -8,3 +8,45 @@
...
+
+  for (form, prof) in vitro_data
+    @series begin
+      seriestype --> :scatter
+      label --> "Original " * "$(form)"
+      prof.time, prof.conc
+    end
+  end
+
+  for (form, prof) in vitro_data
```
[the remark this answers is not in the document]
sipah00: yeah, I can do it

src/ivivc/main.jl
sipah00: changed to `ivivc_predict`

ChrisRackauckas:
Still not completely happy about the interface, but it's much improved. Let's keep continuing this.

PR #645, opened 2019-08-17
Unweighted residuals and re-factoring dv distributions

pkofod:
This calculates the unweighted residuals and moves the `derived_dv_dist` functionality (bikeshedding invitation!) into its own function. One advantage (though marginal in importance) is that there's it's no longer needed to call `conditional_nll_ext` and evaluate `_lpdf` in weighted residuals etc. Useful for simple estimation procedures without etas and as a helper function in the weighted residual calculations.


src/estimation/likelihoods.jl
```diff
@@ -107,14 +107,31 @@ conditional_nll(m::PumasModel,
...
 
     # extract distributions
     derived_dist = m.derived(collated, solution, obstimes, subject)
   end
+  derived_dist
+end
+
 
-  ll = _lpdf(derived_dist, subject.observations)::numtype(collated)
+@inline function conditional_nll_ext(m::PumasModel,
```
andreasnoack: Do we need `conditional_nll_ext` now that we have a separate function for extracting the distributions?

pkofod: Not if we allow for the `dist` to be input into `condtional_nll`

pkofod:
Those pesky `args...` :) I think it's there now, even if it took me quite a while to fix everything :) 

Can you have a look @andreasnoack ?

src/estimation/diagnostics.jl
```diff
@@ -1,3 +1,13 @@
+function StatsBase.residuals(fpm::FittedPumasModel)
+  # Return the residuals
+  return [residuals(fpm.model, subject, fpm.param, vrandeffs) for (subject, vrandeffs) in zip(fpm.data, fpm.vvrandeffs)]
+end
+function StatsBase.residuals(model::PumasModel, subject, param, vrandeffs)
+  # Calculated the dependent variable distribution objects
+  dist = derived_dv_dist(model, subject, param, (η=vrandeffs,))
```
andreasnoack:
```suggestion
  rtrf = totransform(model.random(param))
  randeffs = TransformVariables.transform(rtrf, vrandeffs)
  dist = derived_dv_dist(model, subject, param, randeffs)
```
and similarly for all the other occurrences

src/estimation/diagnostics.jl
```diff
@@ -119,12 +129,12 @@ function wres(m::PumasModel,
...
-  y = subject.observations.dv
-  nl, dist = conditional_nll_ext(m,subject,param, (η=vrandeffs,))
+
+  dist = derived_dv_dist(m, subject, param, (η=vrandeffs,))
   Ω = Matrix(param.Ω)
-  F = ForwardDiff.jacobian(s -> mean.(conditional_nll_ext(m, subject, param, (η=s,))[2].dv), vrandeffs)
+  F = ForwardDiff.jacobian(s -> mean.(derived_dv_dist(m, subject, param, (η=s,)).dv), vrandeffs)
   V = Symmetric(F*Ω*F' + Diagonal(var.(dist.dv)))
-  return cholesky(V).U'\(y .- mean.(dist.dv))
+  return cholesky(V).U'\residuals(m, subject, param, vrandeffs)
```
andreasnoack: This recomputes the distributions, i.e. it solved the dynamical system again even though you've already done that when computing `dist`. Maybe introduce a helper `residual(...,dist,...)` method to avoid that or keep the existing version.

Same comment applies to the versions below.

src/estimation/likelihoods.jl
```diff
@@ -63,31 +63,18 @@ end
...
-                                     args...;
-                                     # This is the only entry point to the ODE solver for the estimation code
-                                     # so we need to make sure to set default tolerances here if they haven't
-                                     # been set elsewhere.
-                                     reltol=DEFAULT_RELTOL,
-                                     abstol=DEFAULT_ABSTOL,
-                                     kwargs...)
-
+# inlined beacuse it was taken from conditional_nll_ext below that is inlined
+@inline function derived_dv_dist(m,
```
andreasnoack: Please add type signatures

src/estimation/likelihoods.jl
```diff
@@ -104,18 +91,49 @@ conditional_nll(m::PumasModel,
...
+                                 subject::Subject,
+                                 param::NamedTuple,
+                                 randeffs::NamedTuple,
+                                 args...;
+                                 kwargs...)
+    derived_dist = derived_dv_dist(m, subject, param, randeffs, args...; kwargs...)
+    conditional_nll(m, subject, param, randeffs, derived_dist)
+end
+
+@inline function conditional_nll(m::PumasModel,
```
andreasnoack: `m` isn't used here so it can be dropped

src/estimation/likelihoods.jl
```diff
@@ -651,13 +670,19 @@ function _conditional_nll_ext_η_gradient(m::PumasModel,
   # Costruct closure for calling conditional_nll_ext as a function
   # of a random effects vector. This makes it possible for ForwardDiff's
   # tagging system to work properly
-  _conditional_nll_ext =  _η -> conditional_nll_ext(m, subject, param, (η=_η,), args...; kwargs...)
+  _derived_dv_dist =  _η -> derived_dv_dist(m, subject, param, (η=_η,), args...; kwargs...)
+  _conditional_nll =  _η -> conditional_nll(m, subject, param, (η=_η,), _derived_dv_dist(_η))
```
andreasnoack: This will solve the system twice. Just keep the first version and call `conditional_nll` after `_conditional_nll_ext_η_gradient` in the methods that use it.

pkofod: This is the only thing I'm not quite sure about (and I was actually going to ask you what should be done here). What are you suggesting to be done exactly? The function is called one place: 20 lines below.

pkofod:
I had fixed most of your comments @andreasnoack but I ended up in a weird state, so I force pushed the commit you commented on. Will try again :)

pkofod:
Some things are a bit different, but now everything works, and  `derived_dist` can be called on its own. `residuals` is used in `wresiduals` when the raw residual appears.

PR #646, opened 2019-08-21
Fix missing Random in Project.toml and remove use of RealVector to ma…

pkofod:
…tch TransformVariables 0.3.5.

Edit: it appears there was more to the transformvar changes...


pkofod:
I was working on an unrelated thing, so in interesting of time I'll upper bound TransformVariables for now. The new version has killed our dispatches.

PR #647, opened 2019-08-25
Add functions for gsa

Vaibhavdixit02:
This PR adds the Global Sensitivity Analysis Routines we have in https://github.com/JuliaDiffEq/DiffEqSensitivity.jl for `PumasModel`s. Starting out with Morris and Sobol, two of the most frequently used GSA methods. More methods will be added as per the ACoP poster submission from later PRs.


src/Pumas.jl
```diff
@@ -33,6 +33,8 @@ include("estimation/bayes.jl")
 include("estimation/diagnostics.jl")
 include("estimation/vpc.jl")
 include("estimation/show.jl")
+include("estimation/saem.jl")
```
ChrisRackauckas: oops?

src/estimation/gsa.jl
```diff
@@ -0,0 +1,27 @@
+using DiffEqSensitivity
+
+abstract type GSAMethod end
```
ChrisRackauckas: IMO these types should live in DiffEqSensitivity. I've been meaning to say we need to fix up that interface.

src/estimation/gsa.jl
```diff
@@ -0,0 +1,27 @@
+using DiffEqSensitivity
```
ChrisRackauckas: top level

src/Pumas.jl
Vaibhavdixit02: Yes

pkofod:
I know you've surely co-ordinated with either Vijay or Chris on this, but for the sake of the rest of us being able to follow, can I ask you to provide a slightly more informative initial message than "No description provided." :)

ChrisRackauckas:
http://docs.juliadiffeq.org/latest/analysis/global_sensitivity.html

ChrisRackauckas:
Tagging DiffEqSensitivity today to test this.

ChrisRackauckas:
I feel like we can do something like @pkofod 's PR for handling the parameter ranges here, but we can merge for now and handle that detail later.

src/estimation/gsa.jl
```diff
@@ -0,0 +1,41 @@
+function DiffEqSensitivity.gsa(m::PumasModel,subject::Subject,params::NamedTuple,method::DiffEqSensitivity.Sobol,p_range=[[0.0,1.0] for i in 1:length(TransformVariables.inverse(toidentitytransform(m.param),params))],args...;order=[0,1],samples=1000,kwargs...)
```
ChrisRackauckas: Can't this all be one method?

Vaibhavdixit02: I am not sure it would work, for Morris you need an additional positional argument of `p_steps` which specifies the number of steps for each parameter.

ChrisRackauckas: I think `p_steps` should be in the definition of Morris, so that way the kwargs are the same for all of them.

Issue #602, opened 2019-07-27
Optim is sensitive to the source of ODE solutions


vjd:
This is linked to #600 For the model given below, the default `alg=AutoTsit5(Rosenbrock23())`  is not stable during optimization.  Switching to `alg=Rosenbrock23()` allows it to converge. Needs an evaluation of the algorithm.
dataset - [SCM_Mutaz.xlsx](https://github.com/PumasAI/Pumas.jl/files/3438258/SCM_Mutaz.xlsx)

```julia
using Pumas, CSV

data = CSV.read("SCM_Mutaz.csv")

data = read_pumas(data, cvs =[:WT],
        dvs = [:dv], id=:ID, time=:TIME,
        evid=:EVID,amt=:AMT,cmt=:CMT,event_data=true)

mod = @model begin
  @param   begin
    tvcl ∈ RealDomain(lower=0, init = 4)
    tvv ∈ RealDomain(lower=0, init = 10)
    tvka ∈ RealDomain(lower = 0, init= 1.5)
    Ω ∈ PDiagDomain(init=[0.09,0.09,0.09])
    σ_prop ∈ RealDomain(lower=0,init=0.02)
  end

  @random begin
    η ~ MvNormal(Ω)
  end
  @covariates WT
  @pre begin
    CL = tvcl  * exp(η[1])
    V  = tvv   * exp(η[2])
    Ka = tvka * exp(η[3])
  end

  #@dynamics OneCompartmentModel
  @dynamics begin
    Depot' =  -Ka*Depot
    Central' =  Ka*Depot - (CL/V)*Central
  end

  @derived begin
      cp = @. (Central / V)
      dv ~ @. Normal(cp, sqrt(cp^2*σ_prop))
    end
end


param = init_param(mod)
f1 = fit(mod, data, param, Pumas.FOCEI())

```

ChrisRackauckas:
Yes, we need to checking the switching parameters and make it a little more sensitive.

YingboMa:
I am not sure if that is a problem with the switching algorithm. Using `TRBDF2` or `Rodas5` which are L-stable algorithms fail to converge, too.

YingboMa:
Every solve was completed successfully, but Optim was not able to converge.
```julia
 ┌ Warning: Linesearch failed, using alpha = 0.0 and exiting optimization.
│ The linesearch exited with message:
│ Linesearch failed to converge, reached maximum iterations 1000.
[...]

PR #650, opened 2019-08-28
lower default tolerance for estimation and use a different solver

ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/602 . By using a lower tolerance, the line search tends to succeed. However, with these tolerances it enters a different performance regime. In this regime, Rodas5 is stable enough and will be faster than Rosenbrock23, while Vern9 is probably the fastest non-stiff method but has higher compile times, so Vern7 is a good middle ground. 

@vjd can you test these new defaults on some examples that were failing?

### DIFF
--- a/src/estimation/likelihoods.jl
+++ b/src/estimation/likelihoods.jl
@@ -614,18 +614,18 @@ function marginal_nll_gradient!(g::AbstractVector,
   return g
 end
 
-function _fdrelstep(model::PumasModel, param::NamedTuple, reltol::AbstractFloat, ::Val{:forward})
+function _fdrelstep(model::PumasModel, param::NamedTuple, reltol, ::Val{:forward})
   if model.prob isa ExplicitModel
     return sqrt(eps(numtype(param)))
   else
-    return max(reltol, sqrt(eps(numtype(param))))
+    return max(norm(reltol), sqrt(eps(numtype(param))))
   end
 end
-function _fdrelstep(model::PumasModel, param::NamedTuple, reltol::AbstractFloat, ::Val{:central})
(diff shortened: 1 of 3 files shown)


ChrisRackauckas:
From Vijay's tests we know this is at least better, and from the discussions we know this is theoretically better, so I am going to merge as it's a clear improvement. We can continue testing to see if it fixes all of the issues, but even then having this done should isolate any remaining inference issues to something else.

pkofod:
Yup, looks good to me as well. We might want to look at some different methods and see how AD'ing through them works as there might be a compile time trade off (I think the new version was slightly slower than the previous default?), but again, robustness wins over compile time unless it's extreme imo.

ChrisRackauckas:
>I think the new version was slightly slower than the previous default?

Yeah slightly, we are `@..`'ing some larger expressions so the code generation is a bit more.

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/verner_rk_perform_step.jl#L246

The fastest method in this case is most definitely Vern9, but 

https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/blob/master/src/perform_step/verner_rk_perform_step.jl#L563

the compiler really doesn't like this line. We can see whether it's acceptable though, it's just a small swap to `AutoVern9` instead. For long optimizations it's 100% what you'd want to run.

ChrisRackauckas:
If someone really wants to go deep, you can test https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/904 and see if that improves timing at all in this case.

PR #653, opened 2019-09-02
Full function building at compile time and test distributed

ChrisRackauckas:


src/dsl/model_macro.jl
```diff
@@ -289,42 +289,92 @@ end
...
-      ___D = Differential(t)
-      $params
-      $bvars
+      $fname = $f_ex
       $diffeq
     end
   end
 end
+
+function convert_rhs_to_Expression(s::Symbol, dvars, params, t)
```
ChrisRackauckas: It might be good to upstream this to MT.

ChrisRackauckas:
@HarrisonGrodin @YingboMa take note that this is what I meant by compile-time usage. 

@andreasnoack @vjd this should fix the distributed issues, and a test was added for that.

src/dsl/model_macro.jl
```diff
@@ -289,42 +289,92 @@ end
...
+  j = findfirst(params) do x
+    if x isa Operation
+      x.op.name == ex.args[1]
+    else # For CL(t)
+      x.name == ex.args[1]
+    end
+  end
+
+  if i === j === nothing
+    op = eval(ex.args[1])  # HACK
```
andreasnoack: Any chance that we can avoid more `eval`ing in macros? If not now then maybe later?

ChrisRackauckas: what's your suggestion for how to handle this?

ChrisRackauckas: found a way.

test/core/dsl.jl
```diff
@@ -38,6 +38,7 @@ mdsl = @model begin
 
     @vars begin
       conc = Central / V
+      conc2 = Central^2
```
ChrisRackauckas: made this test a little harder.

ChrisRackauckas:
It was just a subtle issue with processes in a module, so a general Julia issue. Once that was avoided, tests worked fine. The result is no evals and distributed works in all cases.

pkofod:
Strange, but great you figured it out. Was there already an issue for this?

ChrisRackauckas:
Not sure if there's an issue, but it's really just because `@everywhere` and such hard code their module to be Main

PR #660, opened 2019-09-07
Allow for fixing parameters when fitting.

pkofod:
This is going to be useful for ignoring `eta`s during fitting (because the user will have to specify what parameters control the variance/covariances of the etas), but is also generally useful.

We have to store the new `TransformTuple` or else the printing will be incorrect in `show`. Should we visually mark the fixed parameters in `show`? A column called `fixed` with stars or just an post-fix identifier with a name and a footnote .

I guess we should skip the fixed parameters in things like `vcov`/standard-errors, etc? We "can" calculate them, but they will not be estimated, so the operations will be meaningless.

```julia
julia> o = fit(theopmodel_analytical_fo, theopp, param, Pumas.FO(); fixedparam=(θ₁=4.1, θ₃=0.03))
FittedPumasModel

Likelihood approximation:           Pumas.FO
Objective function value:          158.37124
Total number of observation records:     132
Number of active observation records:    132
Number of subjects:                       12

----------------------
          Estimate
----------------------
θ₁         4.1
θ₂         0.064313
θ₃         0.03
θ₄         2.1444
Ω₁,₁      14.84
Ω₂,₁      -0.029677
Ω₃,₁      -0.97181
Ω₂,₂       0.00023506
Ω₃,₂       0.010032
Ω₃,₃       0.49577
σ_add      0.27107
----------------------


julia> o = fit(theopmodel_analytical_fo, theopp, param, Pumas.FO(); fixedparam=(θ₁=4.1, θ₃=0.03, σ_add=0.1))
FittedPumasModel

Likelihood approximation:           Pumas.FO
Objective function value:          192.59738
Total number of observation records:     132
Number of active observation records:    132
Number of subjects:                       12

----------------------
          Estimate
----------------------
θ₁         4.1
θ₂         0.065205
θ₃         0.03
θ₄         2.1326
Ω₁,₁      14.736
Ω₂,₁      -0.027657
Ω₃,₁      -0.94227
Ω₂,₂       0.00025411
Ω₃,₂       0.010416
Ω₃,₃       0.50645
σ_add      0.1
----------------------
```


src/estimation/likelihoods.jl
```diff
@@ -1046,14 +1046,15 @@ _∂²l∂η²(dv_d::Any,
         args...; kwargs...) = throw(ArgumentError("Distribution is current not supported for the $approx approximation. Please consider a different likelihood approximation."))
 
 # Fitting methods
-struct FittedPumasModel{T1<:PumasModel,T2<:Population,T3,T4<:LikelihoodApproximation, T5, T6, T7}
+struct FittedPumasModel{T1<:PumasModel,T2<:Population,T3,T4<:LikelihoodApproximation, T5, T6, T7,T8}
```
ChrisRackauckas:
```suggestion
struct FittedPumasModel{T1<:PumasModel,T2<:Population,T3,T4<:LikelihoodApproximation, T5, T6, T7, T8}
```

pkofod:
I should say that I considered just having `fixedparam` be a tuple of symbols, and then just fix the values in the input `param`. I think that has two drawbacks: 1) users will have to write `:θ₁` instead of `θ₁` and we will then have to explain why/what/when you put the colon when working with parameters and 2) it also means that you have to re-enter all entries in `param` again every time you want to try to fix a parameter at some value, because there is no easy way for users to just change one entry in a NamedTuple.

pkofod:
> LGTM. Surprised it was that simple.

To be completely honest: me too :)

andreasnoack:
We might want to consider a different name since "fixed" already has meaning in these models.

ChrisRackauckas:
That is a good point that I didn't see. Open an issue to discuss more? Constant?

pkofod:
constant would be in line with the domain and transformation names

vjd:
I think `fix` as a verb is better? Andreas, what `fixed` are you talking about, isn't it about fixing the parameter?

ChrisRackauckas:
fixed effects

pkofod:
we can discuss it here https://github.com/PumasAI/Pumas.jl/pull/661

PR #661, opened 2019-09-08
fixedcoef -> constantcoef

pkofod:
Fixedcoef might have been a bad choice of name in a population model context as pointed out by @andreasnoack . Constant is consistent with `ConstantDomain` and `ConstantTransformation`.

### DIFF
--- a/src/estimation/likelihoods.jl
+++ b/src/estimation/likelihoods.jl
@@ -1115,13 +1115,13 @@ function Distributions.fit(m::PumasModel,
                            # in zero. In addition, the returned object should support a opt_minimizer method
                            # that returns the optimized parameters.
                            optimize_fn = DEFAULT_OPTIMIZE_FN,
-                           fixedcoef = NamedTuple(),
+                           constantcoef = NamedTuple(),
                            kwargs...)
 
   # Compute transform object defining the transformations from NamedTuple to Vector while applying any parameter restrictions and apply the transformations
   trf = totransform(m.param)
   fixedtrf=trf
-  param, fixedtrf = _fixed_to_constanttransform(trf, param, fixedcoef)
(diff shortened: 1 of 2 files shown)


pkofod:
@vjd mentioned `fix` as an alternative in #660 - happy to discuss here

pkofod:
So apparently it's called `ConstDomain` but `ConstantTransform`:) so the question is whether we call it `constcoef` (`const_coef)` or `constantcoef` (`constant_coef`)...

vjd:
`contstantcoef` is fine too. All other software use either `fix` or `freeze`, so thought we may want to think in those lines

pkofod:
I think `fix` could be okay with me as well.

vjd:
being verbose/explicit help, so, if we go with your proposals, then I would suggest `constantcoef` with (`constant`) spelled out fully

vjd:
between `constantcoef` and `fix`, I vote for with `fix`. 

The example tests don't provide this information, but can we also `fix` `omega`'s?

pkofod:
> The example tests don't provide this information, but can we also `fix` `omega`'s?

Yes, I can add a test, but of course you have to specify what the "omegas" are (since they could be called anything)

andreasnoack:
My gut feeling is that `fix` is sufficiently distinct from `fixed` to avoid confusing so I'm fine with that although I slightly prefer `constant`.

We should make `Domain` and `Transform` consistent with respect to `Const` and `Constant`. I prefer `Constant` but `ConstDomain` is user facing so it would be more breaking to change `ConstDomain`. It can be deprecated, though, so we would avoid hard fails and the deprecations could be completed by ACoP.

vjd:
sounds good! I am going to merge this for now.

PR #662, opened 2019-09-09
Simplify PumasModel dispatches

Vaibhavdixit02:
Updates the `gsa` methods for `PumasModel` to uniform interfaces for both Sobol and Morris and match the changes in https://github.com/JuliaDiffEq/DiffEqSensitivity.jl/pull/95


src/estimation/gsa.jl
```diff
@@ -1,41 +1,19 @@
-function DiffEqSensitivity.gsa(m::PumasModel,subject::Subject,params::NamedTuple,method::DiffEqSensitivity.Sobol,p_range=[[0.0,1.0] for i in 1:length(TransformVariables.inverse(toidentitytransform(m.param),params))],args...;order=[0,1],samples=1000,kwargs...)
+function DiffEqSensitivity.gsa(m::PumasModel,subject::Subject,params::NamedTuple,method::DiffEqSensitivity.GSAMethod,p_range=[[0.0,1.0] for i in 1:length(TransformVariables.inverse(toidentitytransform(m.param),params))])
```
ChrisRackauckas: the args... kwargs... should be passing into `simobs`.

PR #666, opened 2019-09-10
Estimate naive method (ignoring etas) in models with random blocks

pkofod:
This allows for zeroing out (and this covers all of our models because the randoms always have zero mean) the random effects and estimating simple methods on models that have random blocks in them.

Now, I am setting `constantcoef=[omega=[0.0],]` in the tests, but technically you don't *have* to do that. The reason is, that all the gradients related to the random effect variance parameters will be 0.0, because they don't affect the likelihood, so they original parameter guess will never be changed. However, I do think it's good to tell users to specify these parameters regardless. If they don't, it's hard (impossible) to know in the post-processing what is what. We may want to discuss if users should explicitly say "these are the "omegas"", not just use the constantcoef interface (we can then use the constantcoef interface internally ourselves...) Maybe the follow example shows what I mean:
```
julia> fitone = fit(mdsl1, data, param, Pumas.NaivePooled())
FittedPumasModel

Likelihood approximation:  Pumas.NaivePooled
Objective function value:          45.496706
Total number of observation records:      20
Number of active observation records:     20
Number of subjects:                       10

------------------
         Estimate
------------------
θ₁        0.36186
Ω₁,₁      0.04
Σ         0.1
------------------


julia> fitone = fit(mdsl1, data, param, Pumas.NaivePooled(); constantcoef=(Ω=[0.],))
FittedPumasModel

Likelihood approximation:  Pumas.NaivePooled
Objective function value:          45.496706
Total number of observation records:      20
Number of active observation records:     20
Number of subjects:                       10

----------------
       Estimate
----------------
θ₁      0.36186
Ω₁      0.0
Σ       0.1
----------------
```
both things work, but only in the latter will we know that omega was not supposed to be optimized over.


pkofod:
It's not super satisfying that people need to fix parameters at random values. I could just as well have put `constantcoef=(Ω=[123123.0])` above, and the other estimates and objective value would have been exactly the same. I don't know what the alternatives are? We could just not have them pass anything as I said before, and the gradient based methods will just not move the omega parameters. We could also introduce a `omegas=(:Ω,)` keyword, but that is *also* not great, as we'd be introducing a new keyword just for this and that adds complexity in terms of teaching and maintanabiliy. The best would be to just be able to infer which parameters were "omegas", but we cannot get that from the current `PumasModel` information. One solution could be to require that omegas were specified in the `random` block instead, but I don't know how popular that would be, and it would also require changes to the macros... Open to suggestions.

vjd:
>I could just as well have put constantcoef=(Ω=[123123.0]) above, and the other estimates and objective value would have been exactly the same. 

I am not sure this is correct. Let us say you have `eta1` on `tvcl` in your model. `eta1` comes from the `omega` distribution. If I `fix` the value of the `omega` to zero, it means that the variance is zero and hence all the sampled `etas` will be zero. But in your example above, if I fix the `omegas` to a large number or any other number for that matter, the `eta1`'s will be sampled from the `omega` distribution with a variance specified by what you fixed to in the `constantcoef`. Is it not the case. I would expect that the parameter estimates will be impacted (in the context of the full variance-covariance matrix). 

If you take another example of just the fixed effects in a simple model like `y=mx+b`, if you `fix` the value of `b` to something arbitrary vs fixing to zero or not using `b` in the model are different things that will impact the estimation of `m`, correct?

Or, did I totally miss the point?

pkofod:
> Or, did I totally miss the point?

I think you missed the context. 

The context is that of: a) the model has a random block and b) you're asking for a simple estimator `NaivePooled`, a single subject or `TwoStage`.

This PR makes sure that if we're in that case, then the random block is effectively ignored by not evaluating the density of the etas in the likelihood expression (in the limit of the variance going to zero, this term becomes a degenerate distribution with a mass point at 0, so we can just set the etas to zero for the model solution part and ignore the density for the likelihood part). The "*problem*" is that we're not even evaluating that density, so even if the variance-covariance matrix is actually fixed at something else, this won't matter (again, we're *not* evaluating anything that omega impacts).

As a matter of teaching users, it might be sufficient to just tell them to set it to 0, but I would really prefer to not have them set a value that can't have any other value. In this example
```
fitone = fit(mdsl1, data, param, Pumas.NaivePooled(); constantcoef=(Ω=[0.],))
```
I'm not using the fact that that variance is set to 0.0, I'm **only** using the fact that the user tells me that the parameter Ω is what affects the random block.

pkofod:
> I'm not using the fact that that variance is set to 0.0, I'm **only** using the fact that the user tells me that the parameter Ω is what affects the random block.

This is why I was wondering if we should use another mechanism than `constantcoef`. One possibility, as I said, is to say

```
fitone = fit(mdsl1, data, param, Pumas.NaivePooled(); omegas=(:Ω,))
```
and we can then fix it to 0 ourselves internally. From a user point of view it's slightly better because they don't have to set a value to something that has no effect because we're not actually setting the variance to zero, we're actually effectively removing the etas all together.

What I could have wished for in this context is that the model DSL looked something like
```
theopmodel_analytical_fo = @model begin
    @param begin
      θ₁ ∈ RealDomain(lower=0.1,    upper=5.0, init=2.77)
      θ₂ ∈ RealDomain(lower=0.0008, upper=0.5, init=0.0781)
      θ₃ ∈ RealDomain(lower=0.004,  upper=0.9, init=0.0363)
      θ₄ ∈ RealDomain(lower=0.1,    upper=5.0, init=1.5)
      Ω ∈ PSDDomain(3)
      σ_add ∈ RealDomain(lower=0.001, init=0.388)
      #σ_prop ∈ RealDomain(init=0.3)
    end

    @random begin
     @from_param Ω # here's a list of parameters from the param block we're allowed to use
      η ~ MvNormal(Ω)
    end

    @pre begin
      Ka = SEX == 0 ? θ₁ + η[1] : θ₄ + η[1]
      K  = θ₂+ η[2]
      CL = θ₃*WT + η[3]
      V  = CL/K
      SC = CL/K/WT
    end

    @covariates SEX WT

    @vars begin
      conc = Central / SC
    end

    @dynamics OneCompartmentModel

    @derived begin
      dv ~ @. Normal(conc,sqrt(σ_add))
    end
  end
```
or
```
theopmodel_analytical_fo = @model begin
    @param begin
      θ₁ ∈ RealDomain(lower=0.1,    upper=5.0, init=2.77)
      θ₂ ∈ RealDomain(lower=0.0008, upper=0.5, init=0.0781)
      θ₃ ∈ RealDomain(lower=0.004,  upper=0.9, init=0.0363)
      θ₄ ∈ RealDomain(lower=0.1,    upper=5.0, init=1.5)
      # no omega here!!
      σ_add ∈ RealDomain(lower=0.001, init=0.388)
      #σ_prop ∈ RealDomain(init=0.3)
    end

    @random begin
     @omegas begin # a separate param block in random that will be merged with the other one internally
       # but the fact that it's defined here, means that we know that this is an omega  
       Ω ∈ PSDDomain(3)
      end
      η ~ MvNormal(Ω)
    end

    @pre begin
      Ka = SEX == 0 ? θ₁ + η[1] : θ₄ + η[1]
      K  = θ₂+ η[2]
      CL = θ₃*WT + η[3]
      V  = CL/K
      SC = CL/K/WT
    end

    @covariates SEX WT

    @vars begin
      conc = Central / SC
    end

    @dynamics OneCompartmentModel

    @derived begin
      dv ~ @. Normal(conc,sqrt(σ_add))
    end
  end
```

vjd:
Thank you for clarifying. I get it now. 
>and we can then fix it to 0 ourselves internally.

I think the specification of the `omegas` by the user and we working on setting it internally to zero is the best way in my opinion.

vjd:
@pkofod let's go with the `kwarg` suggestion and merge this in?

pkofod:
> @pkofod let's go with the `kwarg` suggestion and merge this in?

I'll take this up to date and add the `omega` (`omegas`) keyword.

vjd:
bump

pkofod:
Pushed some changes @vjd . `fit` should work now, but infer doesn't work yet.

vjd:
@pkofod so the blocker on this currently is `infer`?

pkofod:
Yes, but I am looking into it.

PR #674, opened 2019-09-11
Fix NLME test sets that didn't run unless ALL was specified.

pkofod:
Because I had failed to specify their groups in the gitlab config file.


andreasnoack:
Don't merge

ChrisRackauckas:
what's wrong with this?

pkofod:
> what's wrong with this?

Andreas is going to push a commit that changes something in the outer runtest.jl file.

Issue #595, opened 2019-07-19
simobs populations as an ensemble


ChrisRackauckas:
https://github.com/PumasAI/Pumas.jl/blob/master/src/models/model_api.jl#L182-L192 should be making use of the new ensemble interface, which really just requires moving the solve step out from the prob construction.

PR #675, opened 2019-09-11
Use Ensemble interface for simulation

ChrisRackauckas:
This sets up Pumas to use http://docs.juliadiffeq.org/latest/features/ensemble.html for the population simulations, giving it the parallelism. As an added bonus, this fixes the missing `SplitThreads`. Right now I've only done `simobs`. @logankilpatrick can you similarly handle `solve`? Solve should return the ensemble itself, so output handling shouldn't be necessary. Similar input handling is needed. In fact, the extra setup computations can be eliminated by probably grabbing the values out of `sol.prob.p`, but I haven't taken the time to look into it (I know `col` is in there, and so the easiest fix might be to just stick `obstimes` in there as well).


ChrisRackauckas:
BTW, GPU won't work yet because `EnsembleGPUArray` cannot handle callbacks. Thus that's the next thing on my plate.

ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/595

PR #676, opened 2019-09-11
Add no events handling

YingboMa:
I am not sure where should I add the test for this.

@ChrisRackauckas I made this PR, because I am not aware of a way to solve the ODE system without having an event.

### DIFF
--- a/src/simulate_methods/diffeqs.jl
+++ b/src/simulate_methods/diffeqs.jl
@@ -54,7 +54,7 @@ function build_pkpd_problem(_prob::DiffEqBase.AbstractJumpProblem,set_parameters
 end
 
 function ith_subject_cb(p,datai::Subject,u0,t0,ProbType,saveat,save_discont,continuity)
-  isempty(datai.events) && return t0, nothing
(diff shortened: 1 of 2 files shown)


src/simulate_methods/diffeqs.jl
```diff
@@ -54,6 +54,7 @@ function build_pkpd_problem(_prob::DiffEqBase.AbstractJumpProblem,set_parameters
 end
 
 function ith_subject_cb(p,datai::Subject,u0,t0,ProbType,saveat,save_discont,continuity)
+  isempty(datai.events) && return t0, nothing
```
ChrisRackauckas: does `t0` need to be in tstops?

YingboMa: I am not sure what the default should be, so I used `t0`, hoping that it will have minimal runtime impact.

YingboMa:
@ChrisRackauckas where should I add the test for this?

ChrisRackauckas:
https://github.com/PumasAI/Pumas.jl/blob/master/test/core/multiresponses.jl#L32-L100 with a version without events.

Issue #680, opened 2019-09-12
simobs ignores observation time vector from Subject/Population


vjd:
Here is a simple example
The idea below is to generate random samples per subject while creating the population. Although the population is correctly generated with each subject have 4 random samples between 0 and 24 hours, the call to `simobs` ignores this and sticks to its default of simulating every one hour. 

```
using Pumas

po_1cmt = @model begin
    @param   begin
        tvka ∈ RealDomain(lower=0)
        tvcl ∈ RealDomain(lower=0)
        tvvc ∈ RealDomain(lower=0)
        Ω ∈ PDiagDomain(3)
        σ_prop ∈ RealDomain(lower=0)
    end

    @random begin
      η ~ MvNormal(Ω)
    end

    @pre begin
        CL = tvcl * exp(η[1])
        Vc  = tvvc*exp(η[2])
        Ka = tvka * exp(η[3])
    end

    @covariates dosegrp

     @dynamics begin
         Depot' = -Ka*Depot
         Central' =  Ka*Depot - (CL/Vc)*Central
     end

    @derived begin
        cp = @. 1000*(Central/Vc)
        dv ~ @. Normal(cp, sqrt(cp^2*σ_prop))
    end
end

po_lin_param_1cmt = (tvka = 1,
                tvcl = 4.0,
                tvvc = 70.0,
                Ω = Diagonal([0.09,0.09,0.09]),
                σ_prop = 0.04)

ev = DosageRegimen(100)
pop = Population(map(i -> Subject(id=i, evs=ev, cvs=(dosegrp=10,), time=sort(round.(rand(Uniform(0,24),4);digits=2))),1:3))
DataFrame(pop)
obs = simobs(po_1cmt,pop,po_lin_param_1cmt)
DataFrame(obs)
```

ChrisRackauckas:
https://github.com/PumasAI/Pumas.jl/blob/master/src/models/model_api.jl#L58-L60

That case is checking "if it's only events, default to just a nice saving pattern". Otherwise it would only save at the event times which is usually not what you want. Maybe it needs to check for another detail in there.

PR #684, opened 2019-09-13
Modify the precedence of computing `observationtimes`

YingboMa:
Fix https://github.com/PumasAI/Pumas.jl/issues/680

### DIFF
--- a/src/models/model_api.jl
+++ b/src/models/model_api.jl
@@ -55,10 +55,14 @@ function timespan(sub::Subject,tspan,saveat)
 end
 
 # Where to save
-observationtimes(sub::Subject) = isnothing(sub.observations) &&
+# `sub.time` has the highest precedence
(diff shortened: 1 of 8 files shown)


src/models/model_api.jl
```diff
@@ -55,10 +55,14 @@ function timespan(sub::Subject,tspan,saveat)
 end
 
 # Where to save
-observationtimes(sub::Subject) = isnothing(sub.observations) &&
+# `sub.time` has the highest precedence
+# then `0:lastdose+24`
+# then `0:24`
+observationtimes(sub::Subject) = !isnothing(sub.time) ? sub.time :
+                                 isnothing(sub.observations) &&
```
ChrisRackauckas: I'm not sure we need to check for observations. Even if there are no observations using the event time is probably a better default.

PR #686, opened 2019-09-15
Fix extraction of parameters from BayesMCMCResult

andreasnoack:
Define Chains constructor for BayesMCMCResult instead of converting to Chains
inside the fit meethod. Use the new Chains method for showing
BayesianMCMCRResult.

Also, change the tests to regression tests that are always run. The new version of the tests would have detected this regression.


src/estimation/bayes.jl
```diff
@@ -194,21 +202,17 @@ _clean_param(x::NamedTuple) = map(_clean_param, x)
 
 # "unzip" the results via a StructArray
 function param_values(b::BayesMCMCResults)
-  param = b.loglik.model.param
-  trf_ident = toidentitytransform(param)
-  chain = Array(b.chain)
-  StructArray([_clean_param(TransformVariables.transform(trf_ident, chain[c,:])) for c in 1:size(chain)[1]])
+  trf  = Pumas.totransform(b.loglik.model.param)
+  vals = TransformVariables.transform.(Ref(trf), b.chain)
+  return StructArrays.fieldarray(vals)
```
ChrisRackauckas: IIRC this is our last remaining use of `StructArrays`. Is this needed?

andreasnoack: Indeed. I've removed it for now.

Project.toml
```diff
@@ -52,9 +51,7 @@ LabelledArrays = "0.7.1"
 MCMCChains = "0.3.11"
 ModelingToolkit = "0.6"
 Parameters = "0.10, 0.11, 0.12"
-StochasticDiffEq = "6.10"
 TransformVariables = "< 0.3.5"
-julia = "1.1"
```
ChrisRackauckas: why were the bounds removed?

test/nlme/bayes.jl
```diff
@@ -1,11 +1,5 @@
 using Pumas, Test, CSV, Random, Distributions, LinearAlgebra, TransformVariables
 
-if haskey(ENV,"BAYES_N")
```
ChrisRackauckas: I think we should still have an integration test

Project.toml
andreasnoack: Hm. No idea.

andreasnoack:
I found another bug while working on this. The log-determinant of the Jacobian wasn't computed in the transformation of the random effects. I've fixed that and adjusted the test setup to detect such a regression.

In the old CI setup, we never tested with a high `n` and that was some of the reasons why the two regressions fixed here were not detected (the last was that everything was behind `try/catch`. I've changed the tests such that we use a fairly large `n` with the closed form solution and a small `n` when testing with the ODE solver based version. Both cases fix the seed and the former test against the validated solution.

test/nlme/bayes.jl
andreasnoack: See comment below.

src/estimation/transforms.jl
```diff
@@ -284,10 +284,14 @@ totransform(d::MvNormal) = MvNormalTransform(d)
 TransformVariables.dimension(t::MvNormalTransform) = length(t.d)
 function TransformVariables.transform_with(flag::TransformVariables.LogJacFlag, t::MvNormalTransform,
                                            x::TransformVariables.RealVector{T}) where T
+  if !(flag isa TransformVariables.NoLogJac)
+    ℓ = logdet(t.d.Σ)/2
```
ChrisRackauckas: Is this why the Omega tests were failing and now it matches?

andreasnoack: No. We are still not able to match the Omega values for some reason. The fix here is more fundamental. The bug affected all parameters but hadn't been noticed because of the test setup.

andreasnoack:
I'm not sure what is going on here. The tests pass if I `include` the `bayes.jl` file from the REPL but `GROUP="NLME_BAYES" julia --color=yes -e 'using Pkg; Pkg.test("Pumas")'` fails.

src/estimation/bayes.jl
```diff
@@ -160,31 +162,37 @@ function Distributions.fit(
   # Run the MCMC sampler
   samples, stats = sample(h, prop, vparam_aug, nsamples, adaptor, nadapts; progress=true)
 
+  return BayesMCMCResults(bayes, samples, stats)
+end
+
+function Distributions.fit(model::PumasModel, data::Population, param::NamedTuple, ::BayesMCMC,
+  args...; nsamples=5000, kwargs...)
+  fit(model, data, BayesMCMC(), param,args...;nsamples = nsamples, kwargs...)
```
pkofod: Is it a stylistic choice to not just have `approx::BayesMCMC` in the method signature and then input that as `..., data, approx, param, ...` below? I know there are no fields in BayesMCMC so it's essentially equivalent.

andreasnoack: Yes. When using a type just for dispatch, I usually avoid a variable for it. I think it make it more clear what is passed to the next function.

andreasnoack:
This is pretty bananas. The test failure is because of bounds checking when running with `Pkg.test`. Apparently, it has a quite big effect on the sampled values. I'll update the tests based on bounds checking but it means that tests will fail if run without bounds checking.

andreasnoack:
I hadn't realized how sensitive MCMC (maybe just HMC?) is to rounding errors. It was way morere work than expected to get the tests working. I've commented out the MCMC part of ODE solver tests since they are too unstable unless we increase the number iterations by a lot and that will be too expensive. We are testing the same model with a close form solution and we a still testing the log-density functions for the ODE solver based solution so it shouldn't be a big a deal. I'll merge to get the significant Bayesian issues resolved and then we can fine tune later if needed.

ChrisRackauckas:
Check whether lower tolernaces here gives better gradients.

andreasnoack:
I doubt that it was enough. I had to even drop the display test for the closed form solution even though it rounds to six significant digits so I guess this is just a property of HMC.

PR #689, opened 2019-09-17
Fixes to make tests pass when run in serial on new machine.

andreasnoack:
Move all distributed tests to separate directory. Check that Hessian
plus identity is pd before trying to compute the `logdet`.

The release revealed several issues and most of them were releaved just from running the tests in a single group on Arctic1. This was partly because of the single group and partly because of Arctic1 having new processor architecture which triggered some numerical errors.


ChrisRackauckas:
Can you rebase this onto https://github.com/PumasAI/Pumas.jl/pull/688 and see whether this works on Arctic?

andreasnoack:
Yes, I can try that. We should do a sequential run on Arctic1 before each release to avoid that Harsha hits these failures that are unrelated to JuliaPro.

andreasnoack:
> Can you rebase this onto #688 and see whether this works on Arctic?

Tests on Arctic1 just passed

PR #691, opened 2019-09-21
Add covariates to subject show

ChrisRackauckas:


src/data_parsing/data_types.jl
```diff
@@ -393,6 +393,11 @@ function Base.show(io::IO, subject::Subject)
                      observables)
     println(io, "  Observables: $vals")
   end
+  if length(subject.covariates) >= 100:
```
ChrisRackauckas:
```suggestion
  if length(subject.covariates) >= 10:
```

logankilpatrick: done!

ChrisRackauckas:
Tests fail.

src/data_parsing/data_types.jl
```diff
@@ -393,6 +393,13 @@ function Base.show(io::IO, subject::Subject)
                      observables)
     println(io, "  Observables: $vals")
   end
+  if subject.covariates != Nothing
```
asinghvi17: Shouldn't this be `nothing`?

ChrisRackauckas: yes, it should be `!== nothing`

PR #695, opened 2019-09-26
Avoid some inverse conversions of parameters to vectors

andreasnoack:
This is a followup on https://github.com/PumasAI/Pumas.jl/pull/629. In couple of places, we can avoid converting back and forth between the `NamedTuple` and `Vector` representation of the random effects. I should be more efficient and more stable as it removes the last places where we do a linear solve with the covariance matrix of the random effects.


pkofod:
Great! The current version makes it harder to follow the function calls around because you have to think a bit harder if something is a vector or named tuple at all times.

PR #696, opened 2019-09-27
Used NamedTuples for dist to allow for dvs that aren't called `dv`

pkofod:
Had to wrap my head around a few things, and start over to get this done.

I had some more changes earlier, that removed FOCE, but since we're keeping it around for cwres I moved them to another branch.

This doesn't go all the way for multiple dv. You can estimate some things with multiple dvs, but I'm intentionally being vague here, because I'm going to do another "multiple dv" PR that actually properly supports it. This includes not re-calculating `dv0`s multiple times for heteroscedastic models and more. It also includes returning the diagnostics as named tuples.

There's still something we can do to tighten up FOCE (avoid code duplication)


pkofod:
What to do for diagnostics ? Should we just return named tuples or should we rather have users specify the dv by name if there is more than one?

ChrisRackauckas:
>What to do for diagnostics ? Should we just return named tuples

Yes, return named tuples IMO

pkofod:
> Yes, return named tuples IMO

would be consistent with how the dv's are handled in the rest of the code as well. Will it be awkward for single-dv uses?

ChrisRackauckas:
Single dv should always be considered a subset of standard usage without special syntax.

pkofod:
Diagnostics are almost done. The only problem is that some Jacobians are horribly inefficient for multiple dv. However, since we don't advertise multiple dv with this particular PR that might be okay (I just need to figure out how to do it exactly -> I don't want to resolve the same system many times with duals). More testeing needs to be done on multiple dv before we annouce that as "working" though this PR should do much of the work.

I'll remove FOCE in another PR. I wanted to do it here, but there's no need to hold this back.

pkofod:
Okay this works now. As you can see in the tests there is some basic multiple dv support but we need to understand it much better in terms of what features work / performance / etc before we advertise it.

Diagnostics now return named tuples. If any tutorials refer to diagnostics, we need to correct that.

src/estimation/likelihoods.jl
```diff
@@ -288,8 +280,9 @@ function empirical_bayes_dist(m::PumasModel,
 
   parset = m.random(param)
   trf = totransform(parset)
-  _, _, H = ∂²l∂η²(m, subject, param, vrandeffsorth, approx, args...; kwargs...)
-  V = inv(H + I)
+  dv_∂²l∂η² = ∂²l∂η²(m, subject, param, vrandeffsorth, approx, args...; kwargs...)
+  # 3 is W, should we make is a named tuple with l, g, W?
```
andreasnoack: Maybe a struct since the fields will always be the same. I think it would improve the readability quite a bit since it's not obvious what all the `[1]`, `[2]`, and `[3]` refer to.

pkofod:
> Great work. This didn't require as many changes as I'd expected. I'll merge as is and then we can add a struct for the derivatives in a separate PR since it's an internal change.

I will make that PR. Yeah, in the end it ended up being not too bad.

vjd:
Thanks much @pkofod! We are making great progress.

PR #699, opened 2019-10-01
Re-sample etas in npde

pkofod:
Currently we use EBEs in NPDEs which is wrong, they should be resampled together with the dvs.


andreasnoack:
How do we currently test `npde`?

pkofod:
That’s a good question, I just made this one line or to remember it :) i guess the only thing we can test for is standard normality? Might be hard

andreasnoack:
Currently, `npde` is exercised in https://github.com/PumasAI/Pumas.jl/blob/4a9ab0f9d9a943f98aaca60d5de4565ff27f5813/test/nlme/theophylline.jl#L613-L628 and https://github.com/PumasAI/Pumas.jl/blob/4a9ab0f9d9a943f98aaca60d5de4565ff27f5813/test/nlme/simple_model_diagnostics.jl#L37-L53 but the values are not tested against anything. I'll suggest that you fix the seed make the two a regression test for now. Eventually, we should verify that they actually compute something meaningful but it would be good to get the immediate bugfix merged.

pkofod:
>  but the values are not tested against anything. I'll suggest that you fix the seed make the two a regression test for now. Eventually, we should verify that they actually compute something meaningful but it would be good to get the immediate bugfix merged.

Sounds reasonable.

andreasnoack:
bump

pkofod:
had not seen the bump, and forgotten this. but yes we need this in before the next release simply because it's wrong. I'll test against some reference values

pkofod:
I've added some reference values, and I hope they go through on CI. However, I need to free it from the dv name restriction, and I also need to look into why so many output values aren't finite (inf and -inf).

andreasnoack:
Would be good to get this one in.

pkofod:
It turns out the decorrelation of the observed values was done with the full vcov not the "sqrt". This gave some very weird values! Now, I just have to adapt it to the arbitrary dv name changes...

PR #700, opened 2019-10-01
User defined ivivc model and AD support

sipah00:
- [x] User defined IVIVC Model
- [x] AD Support
- [x] An Example


sipah00:
Here is an example of establishing user-defined `ivivc model` which uses `ForwardDiff` pkg for estimating derivatives under the hood.

```julia
using Pumas.IVIVC
using Plots

vitro_data = read_vitro("./vitro_data.csv");
uir_data = read_uir("./uir_data.csv")
vivo_data = read_vivo("./vivo_data.csv");

# user defined ivivc model (quadratic relation for example)
ivivc_model(x, p) = @. p[1] * (x^2) + p[2] * x + p[3]
# Note: This is not exact form of IVIVC Model, it internally modified,
        # and x is being treated as Fdiss(t) and time point t is scaled and shifted according to 
        # keyword args passed to IVIVCModel (see below for more clarification)

p = rand(5)  # remember first two entries are always Tscale and Tshift,
             # so total size of this vector is 2 + number of params in model
lb = fill(-Inf, 5)
ub = fill(Inf, 5)

model = IVIVCModel(vitro_data, uir_data, vivo_data, vitro_model=:emax,
                                                    time_scale = true,
                                                    time_shift = true,
                                                    p = p, lb = lb, ub = ub,
                                                    uir_frac=1.0, deconvo_method=:wn, ivivc_model=ivivc_model);

sol = predict_vivo(model, "medium")
time = vivo_data[1]["medium"].time
conc = vivo_data[1]["medium"].conc
cmax_pe, auc_pe = percentage_prediction_error(time, conc, sol.t, sol.u)
@show cmax_pe, auc_pe;
scatter!(time, conc, label="Original C (medium)")
plot!(sol, label="Predicted C (medium)")
```

vvv_

src/ivivc/main.jl
```diff
@@ -64,24 +69,35 @@ function IVIVCModel(vitro_data, uir_data, vivo_data;
...
+  elseif typeof(ivivc_model) <: Function
+    function m(form, time, p)
+      Tscale = 1.0; Tshift = 0.0
+      if time_scale Tscale = p[1] end
+      if time_shift Tshift = p[2] end
+      ivivc_model(vitro_data[1][form](time*Tscale .- Tshift), p[3:end])
+    end
   else
-    error("Incorrect keyword for IVIVC model!!")
+    error("type of ivivc_model should be Symobol (:two, :three and :four) or Function")
```
ChrisRackauckas:
```suggestion
    error("type of ivivc_model should be Symbol (:two, :three and :four) or Function")
```

src/ivivc/main.jl
```diff
@@ -64,24 +69,35 @@ function IVIVCModel(vitro_data, uir_data, vivo_data;
...
-    p = [0.8, 0.5, 0.6]
-    ub = [1.25, 1.25, 1.25]
-    lb = [0.0, 0.0, 0.0]
-  elseif ivivc_model == :four
-    m = (form, time, x) -> (x[1] * vitro_data[1][form](time * x[2] .- x[3])) .- x[4]
-    p = [0.8, 0.5, 0.6, 0.6]
-    ub = [1.25, 1.25, 1.25, 1.25]
-    lb = [0.0, 0.0, 0.0, 0.0]
+  # IVIVC modeling
+  if typeof(ivivc_model) <: Symbol
```
ChrisRackauckas: No reason for the symbol system. Just have the others be pre-defined functions that can be passed in, and grab the parameters and bounds in that case.

sipah00: sounds good, will do the changes.

sipah00: can you suggest names for these three functions

sipah00:
```julia
using Pumas.IVIVC

using Plots

vitro_data = read_vitro("./vitro_data.csv");
uir_data = read_uir("./uir_data.csv")
vivo_data = read_vivo("./vivo_data.csv");

using Pumas.IVIVC: m_inc # model with intercept

p0, lb, ub = get_ivivc_p0_and_bounds(time_lag=true, time_shift=true, intercept=true)

model = IVIVCModel(vitro_data, uir_data, vivo_data, vitro_model=:emax,
                                                    uir_frac=1.0,
                                                    deconvo_method=:wn,
                                                    ivivc_model=m_inc,
                                                    time_scale=true,
                                                    time_shift=true,
                                                    p=p0, lb=lb, ub=ub);

sol = predict_vivo(model, "medium")
time = vivo_data[1]["medium"].time
conc = vivo_data[1]["medium"].conc
cmax_pe, auc_pe = percentage_prediction_error(time, conc, sol.t, sol.u)
@show cmax_pe, auc_pe;
scatter!(time, conc, label="Original C (medium)")
plot!(sol, label="Predicted C (medium)")
```

sipah00:
@ChrisRackauckas I made a function for p0 and bounds (which can be improved later, better initialization or bounds), if you have a different idea to grab p0 and bounds, please let me know.

src/ivivc/utils.jl
```diff
@@ -146,3 +146,11 @@ function _bateman(conc, time, box, p0, ub, lb)
   end
   lb, ub, p0
 end
+
+function get_ivivc_p0_and_bounds(;time_lag, time_shift, intercept)
```
ChrisRackauckas: why not do this up in the keyword argument handling?

src/ivivc/models.jl
```diff
@@ -81,3 +81,8 @@ e_der(t, p_n) = (t < 0.0) ? 0.0 : @. p_n[1] * p_n[2] * (p_n[3]^p_n[2]) * (t^(p_n
 
 # Weibull model
 w_der(t, p_n) = (t < 0.0) ? 0.0 : @. p_n[1] * (p_n[3]/(p_n[2]^p_n[3])) * (t^(p_n[3]-1)) * exp(-((t/p_n[2])^p_n[3]))
+
+
+# Basic IVIVC Models
+m_w_inc(x, p) = p[1] * x  # model without intercept
```
ChrisRackauckas: I'm not sure these need a pre-definition.

src/ivivc/utils.jl
sipah00: these two params `time_scale` and `time_shift` are already in function key args, but intercept is related to the linear models (defaults) and may not present in a different user-defined model. Our aim to keep linear model ready to use in pkg with some `p0` and `bounds`. 
So, if we include `intercept` as keyword arg and then how will we know that passed ivivc model is default one or user-defined (user-defined model can also have intercept param and in this case, the user will pass p0 and bounds so no need to generate p0 and bounds).
We need something so that we can know that user is passing default model and we need to generate p0 and bounds (if not provided by user) or we can make p0 and bounds keyword args necessary always???

sipah00: OR we can write a bunch of `if-else` to deal with this?

ChrisRackauckas: >Our aim to keep linear model ready to use in pkg with some p0 and bounds.

We don't need to predefine a linear function for a user.

sipah00: Okay, will remove the predefine functions

sipah00:
@ChrisRackauckas Now this can be merged or any other thing you want me to change or add?

ChrisRackauckas:
I didn't see the last commits.

PR #706, opened 2019-10-10
Allow to input new data in predict

pkofod:
This allows users to validate their fit on a hold-out data set or new data obtained later. Very handy to assess external validation.


src/estimation/diagnostics.jl
```diff
@@ -516,24 +516,33 @@ function StatsBase.predict(model::PumasModel, subject::Subject, param, approx, v
   SubjectPrediction(pred, ipred, subject, approx)
 end
 
-function StatsBase.predict(fpm::FittedPumasModel, approx=fpm.approx; nsim=nothing, timegrid=false, newdata=false, useEBEs=true)
+function StatsBase.predict(fpm::FittedPumasModel, approx=fpm.approx; nsim=nothing, timegrid=false, data=nothing, useEBEs=true)
```
andreasnoack: Did you consider doing this with dispatch?

pkofod: I did, yes, the ifelse mess made me think to either have predict dispatch or at least change the internal part.

pkofod: I’ll make variants with fpm and then a population (defaulting to the one in fpm) and then approx. They then map to get subject dispatch at the lowest level

pkofod:
I think this is good to go, but let's wait for the master merge to finish its tests.

PR #708, opened 2019-10-13
WIP: Print license at build time

ChrisRackauckas:
Essentially this would make it so that way every time someone adds or updates Pumas it'll print the license. I need to find out how to make it not go to the log there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.jl.*.cov
 *.jl.mem
 Manifest.toml
+build.log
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Also fails. Maybe @stefankarpinski knows how to do it?

asinghvi17:
https://github.com/JuliaLang/Pkg.jl/blob/f4bdf51d829f98b3cf26a5e4b1de9a21109d3dfa/src/Operations.jl#L831-L835

is what hijacks printing.  Probably not too easy to break out from that, since it captures all output streams.

PR #710, opened 2019-10-14
Koopman and Monte Carlo Expectation calculations

ChrisRackauckas:


ChrisRackauckas:
100x speedup to get 3 digits of accuracy in the probability that AUC<300

test/features/uq_window.jl
```diff
@@ -0,0 +1,57 @@
...
+
+param = (θ = [2.268,74.17,468.6,0.5876],)
+subject1 = data[1]
+obs = simobs(m_diffeq,subject1,param)
+
+function g(obs)
+    obs[:auc] < 300
+end
+param_dists = (θ = Uniform.(m_diffeq.param.params.θ.lower,m_diffeq.param.params.θ.upper),)
+@time uq_windowcost(g,KoopmanQuant(HCubatureJL()),m_diffeq,subject1,param_dists)
```
ChrisRackauckas: @vjd bikeshed the name?

vjd: I was going to ask you the origin of the name! Since this is more specific to uncertainty quantification `uq` over any cost function`g`, I would be in favor of expanding that to just `uncertaintyquantification` or just `uq` and drop the `windowcost`. It just happens to be that we are using a function to compute the probability to be in that `window`. Correct?

ChrisRackauckas: I renamed it all around the fact that, at its core, it's an expectation

PR #712, opened 2019-10-20
Fix _E_and_V for multiple dv and add an _expected_information_fd

andreasnoack:
method that uses finite difference derivatives.


andreasnoack:
Test failure is because of https://github.com/JuliaDiff/ForwardDiff.jl/pull/414

PR #719, opened 2019-10-29
[RFC] Fix VPC

andreasnoack:
The current VPC has some issues. As recently noted, plotting does no longer works with the most recent versions of `Plots`. I believe that is because wee invoked the `ribbon` argument incorrectly which happened to produce a (wrong) result in prior versions of `Plots`. Here is how the current plot in the README looks like
Screen Shot 2019-10-29 at 10 05 03
If you look carefully, you can see that all the four confidence intervals (the two middle ones are almost overlapping) have exactly the same width everywhere which is not intentional.

Furthermore, only the median of the samples is computed and not upper and lower quantiles so the data can be compared to the predictions intervals in the tails of the distribution.

I've changed the implementation such that the same model now will produce
Screen Shot 2019-10-29 at 23 13 24

Two things are remaining but I'll suggest that we'll handle them in separate issues in order to fix the current version which is used in the README and the tutorials. The first thing is that my implementation doesn't handle stratification. The second is that we should find a way to make the empirical quantiles more smooth. It seems that the more common choice is so called "binning" but I'd prefer if we tried some kind of non-parametric regression method instead.


andreasnoack:
I've added (basic) tests and updated the README so I think it's good to go. The plot in the top was based on more subjects to make it easier to see the quantile lines so it's different from the one that will go into the README. It will look like
Screen Shot 2019-10-30 at 09 59 13

vjd:
Thanks for putting in the fixes. The plot above, in my opinion, is the reason we need some sort of binning algorithms on the independent variable as the jaggedness is meaningless and uninterpretable.

src/estimation/vpc.jl
```diff
@@ -1,375 +1,97 @@
...
-struct VPC_QUANT
-  Fiftieth::AbstractVector
-  Fifth_Ninetyfifth::AbstractVector
-  Simulation_Percentiles::AbstractVector
+struct VPC
+  time
+  empirical
+  simulated
+  probabilities
+  confindense_level
```
Vaibhavdixit02:
```suggestion
  confindense_level
```
confidence_level

src/estimation/vpc.jl
```diff
@@ -1,375 +1,97 @@
...
 
-struct VPC
-  vpc_dv::Vector{VPC_DV}
-	Obs_vpc::OBS_VPC
-  Simulations::AbstractVector
-  idv::Symbol
-  data::Population
-end
+  # Simulate `reps` new populations
+  sims = [simobs(m, population, param) for i in 1:reps]
```
Vaibhavdixit02: We should be returning these simulations as well in `VPC`, helps in avoiding repeated simulations (which has much more overhead) in case of switching between stratification.

ChrisRackauckas:
why the removal of safe testsets? Is there a specific function that needs to be shared between the tests? You can use the underlying module for that.

andreasnoack:
> why the removal of safe testsets?

I'd like to use a common dict of test models such that models can be reused in several tests. The temporary modules makes that cumbersome. I'll try to keep the safe testsets and work around it but I'd prefer to get rid of the safe testsets and just rely on the standard test infrastructure.

pkofod:
> Thanks for putting in the fixes. The plot above, in my opinion, is the reason we need some sort of binning algorithms on the independent variable as the jaggedness is meaningless and uninterpretable.

I don't understand why the jaggedness makes it meaningless/uninterpretable. Isn't the visual "test" just if they stay within the conf ints or not?

PR #722, opened 2019-11-04
Improved test coverage for gsa

Vaibhavdixit02:


ChrisRackauckas:
How about 2 derived variables or vector derived variables?

Vaibhavdixit02:
@ChrisRackauckas 2 derived variables does work but the vector derived variables needs fixing which I am working on in https://github.com/PumasAI/Pumas.jl/pull/707, I had a good discussion with Andreas for the same and I will add the tests in that PR once it is fixed.

ChrisRackauckas:
Okay, let's keep pushing one step at a time.

PR #727, opened 2019-11-12
Restrict supported versions of DiffEqDiffTools to avoid breakage

andreasnoack:
The latest minor release of DiffEqDiffTools broke Pumas so we'll have to restrict the supported versions.


ChrisRackauckas:
What's the issue? We'll need to fix this very soon.

PR #731, opened 2019-11-13
Upper bound OrdinaryDiffEq to avoid breakage

andreasnoack:
The 5.20 release of OrdinaryDiffEq is causing our tests to fail so this PR puts an upper bound on the package for the time being.


ChrisRackauckas:
The fix is pretty much in.

andreasnoack:
Sounds good. We can bump the bound here once it's in.

PR #732, opened 2019-11-13
Fix doses such that rates don't reset simultaneous doses and rates if…

pkofod:
… rates come last.

In the current master, a rate=0 doseregimen followed by a rate>0 doseregimen vil overwrite the rate=0 events instead of accumulating them. This is currently handled correct if DiffEq is used, so it's only a fix on the analytical side. @ChrisRackauckas


pkofod:
I'll add a test

PR #733, opened 2019-11-13
Enable the use of DiffEq plotting by defining has_syms and getpropert…

pkofod:
…y for :syms.


ChrisRackauckas:
Can't really test plotting. Plus this is only a dev feature.

andreasnoack:
> Can't really test plotting.

I think it's useful to exercise the code paths even if we can't test that the plot is correct.

pkofod:
I'll test that getproperty actually returns the correct varnames for the different models.

PR #735, opened 2019-11-14
Handle heterogenous time across subjects

Vaibhavdixit02:


andreasnoack:
Generally, the computations here have the structure of grouped dataframe operations (split-apply-combine) and I think it will be much simpler to express the computations in that form. So first step is to convert the data to a DataFrame and then compute quantiles grouped on time as a start. I.e. something like

```julia
julia> df = DataFrame(t=[1, 2, 3, 1, 2], dv=[1.0, 2.0, 1.0, 2.0, 1.0]);
julia> by(df, :t, :dv => t -> quantile(t, [0.1, 0.5, 0.9])')
3×4 DataFrame
│ Row │ t     │ x1      │ x2      │ x3      │
│     │ Int64 │ Float64 │ Float64 │ Float64 │
├─────┼───────┼─────────┼─────────┼─────────┤
│ 1   │ 1     │ 1.1     │ 1.5     │ 1.9     │
│ 2   │ 2     │ 1.1     │ 1.5     │ 1.9     │
│ 3   │ 3     │ 1.0     │ 1.0     │ 1.0     │
```

Then it should hopefully not be too hard to group also on dose size and covariates later on.

Vaibhavdixit02:
vpc

For the readme model this is the vac output

Vaibhavdixit02:
@andreasnoack LGTM?

src/estimation/vpc.jl
```diff
@@ -0,0 +1,159 @@
...
+  stratify_on = nothing
+  )
+
+  time = collect(Iterators.flatten([subject.time for subject in population]))
+
+  # Compute the quantile of the samples
+  empirical, cvs__ = quantile_t_data(population, probabilities, dvname, time, stratify_on)
+
+  # Simulate `reps` new populations
+  sims = [simobs(m, population, param) for i in 1:reps]
```
ChrisRackauckas: this is missing args...;kwargs...

Vaibhavdixit02:
@andreasnoack should be good to go now?

andreasnoack:
You can also reuse the HCV model in https://github.com/PumasAI/Pumas.jl/blob/e5b75ab7a49eb34ca74e277f89b02516a4da1a93/test/nlme/hcv.jl

PR #736, opened 2019-11-14
[Don't merge] Test some stuffs

YingboMa:


ChrisRackauckas:
https://github.com/PumasAI/Pumas.jl/pull/736/commits/c5bc95fa22bbdc1b1af9be158ede8d404dd41adc passed, which means the issue is fastpow, but only on Gitlab...

andreasnoack:
We identified the root cause so I'll close this.

PR #737, opened 2019-11-14
Test getproperty for ExplicitModels.

pkofod:
Tests for #733 . Makes sure that the syms getproperty overload works (returns the right compartment names).

### DIFF
--- a/test/core/analytical_models.jl
+++ b/test/core/analytical_models.jl
@@ -1,0 +1,5 @@
+using Pumas
+
+ImmediateAbsorptionModel().syms == [:Central]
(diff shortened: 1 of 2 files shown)


andreasnoack:
Could you add one for `has_syms` as well?

pkofod:
> Could you add one for `has_syms` as well? I think it would be good to test the full plotting code path, at least eventually, since that is actually what we'd like to have working. If that plotting interface changes at some point these functions might still work but plotting might not.

I agree, but do you want to pull in Plots?  And do you want to do an actual visual regression test?

PR #738, opened 2019-11-17
Fix `lambdaz`'s behavior when the concentration is non-positive

YingboMa:


YingboMa:
PKNCA has a relaxation factor here https://github.com/billdenney/pknca/blob/8f4ed13fd2698fd5e88f64b9d4b119e32b2ef345/R/half.life.R#L209

PR #740, opened 2019-11-18
Analytical models extensions + rewrite

pkofod:
Here it is. Just in time for the 18th :)

So what is in here. These are the analytical solutions that will be used on each interval as defined by the event handling system. I have started writing some explicit tests up against eigen values and vectors derived from some semi-random parameters (I just chose some that were all different and different from one). I will add tests against DiffEq solutions.

I have benchmarked against DiffEq wrt timings and fits. The benchmarks are partial, and I will update and post them, but there seems to be decent time savings, and also precision gains in estimation. There are some issues with `numtype` being calculated over and over again, which becomes the dominating thing in the fits if you have too many "single" parameters specified. The slightly annoying recommendation right now is to use a vector of parameters - but you ofc lose the easy to recognize names here, and have to write everything in terms of indeces.

There's a lot of structure in these models. Adding depots could in principle be made modular as far as I can see, same for adding transit compartments, chains of metabolite dynamics, and so on. This is what ModellingToolkit was supposed to be able to recognize at some point, but I actually think we can derived analytical corrections such that you could in principle write something like

```
@pre begin
  myrate1 = ...
  myrate2 = ...
  myrateN = ...
end
@dynamics NotAdvan
  Depot-> Central | myrate1
  Central -> Output1 | myrate2
  Central <-> Peripheral1 | (myrate3, myrate4)
  Central -> Metabolite1 | myrate5
  Metabolite1 -> Output2 ...
  Metabolite1 -> Metabolite2 ...
  Metabolite2 -> Output3 ..
end
```
and we could build the solution from that - but I think we should save that for a rainy day.

This PR does not implement MM-style analytical solutions. Let's also save those for later / for when someone demands it.

I still need three compartment solutions. I know what they are, but they're a mess to write up. They're important, but I think they can wait.

I need to benchmark existing simple models against this structure to make sure that there are no regressions. I hope static arrays are smart enough but 🤷‍♂


src/analytical_solutions/standard_models.jl
```diff
@@ -1,59 +1,208 @@
...
+    𝕍 = @SMatrix([frac1 0     0;
+                  0     frac2 0;
+                  1     1     1])
+
+    return Λ, 𝕍
+end
+varnames(::Type{OneCmtTwoDepotModel}) = (:Depot1, :Depot2, :Central)
+pk_init(::OneCmtTwoDepotModel) = SLVector(Depot1=0.0,Depot2=0.0,Central=0.0)
+
+function _Λ(::TwoCmtPeriModel, a, b, c)
```
vjd:
```suggestion
function _Λ(::TwoCmtModel, a, b, c)
```

vjd: Two compartment model is implicitly known to have a peripheral compartment, so to be consistent with `OneCmtModel` we can just call it `TwoCmtModel`?

src/analytical_solutions/standard_models.jl
```diff
@@ -1,59 +1,208 @@
...
+    Λ = _Λ(m, a, b, c)
+    𝕍 = @SMatrix([(Λ[1]+c)/b (Λ[2]+c)/b;
+                   1          1        ])
+
+    return Λ, 𝕍
+end
+varnames(::Type{TwoCmtPeriModel}) = (:Central, :Peripheral)
+pk_init(::TwoCmtPeriModel) = SLVector(Central=0.0, Peripheral=0.0)
+
+struct TwoCmtDepotPeriModel <: ExplicitModel end
```
vjd:
```suggestion
struct TwoCmtDepotModel <: ExplicitModel end
```

Same note as above - to be consistent with `OneCmt`

src/analytical_solutions/standard_models.jl
```diff
@@ -1,59 +1,208 @@
...
+  𝕍 = hcat(v_depot, 𝕍)
+
+  return Λ, 𝕍, inv(𝕍)
+end
+varnames(::Type{TwoCmtDepotPeriModel}) = (:Depot, :Central, :Peripheral)
+pk_init(::TwoCmtDepotPeriModel) = SLVector(Depot=0.0, Central=0.0, Peripheral=0.0)
+
+
+# use Vc and Vm
+struct Metabolite011 <: ExplicitModel end # 011?
```
vjd: what is the topology for this one? for metabolites, we may need different parameter names such as `CLm` and `Vm`

vjd: so, from here https://github.com/PumasAI/Pumas.jl/pull/740/files#diff-2baaeea31bb20082f0dd0d07f25e345cR166 I see that it 2 cmt parent and 2 cmt metabolite, correct?

vjd: How about being explicit - `TwoCmtParentTwoCmtMetabolite`
I know it is long, but at least readable and easy to remember?

src/analytical_solutions/standard_models.jl
```diff
@@ -1,59 +1,208 @@
...
+  -(α + sqrt(α^2 + 4*a*c))/2
+end
+function _meta_λ2(a, b, c)
+  α = a + b + c
+  -(α - sqrt(α^2 - 4*a*c))/2
+end
+
+
+# use Vc and Vm
+_Λ(::Metabolite01, a, b, c, d) = _Λ(TwoCmtPeriModel(), a+d, b, c)
```
vjd: and for this -
How about being explicit - `TwoCmtParentOneCmtMetabolite`

pkofod:
I don't know what good names would be. Maybe it should be simpler. Something like.  `Metabolite11` `Metabolite21`, `Metablite12`, `Metabolite22`? For "normal" models: `Central1`, `Cental2`, `Central3` (for three compartment models? Add a depot and it's `DepotMetaboliteXY` and `DepotCentralX` or Transit compartments and it's `TransitCentralX` or `TransitMetaboliteXY`. For the fast and slow absorption models you showed yesterday it could be `FastSlowCentralX` or `FastSLowMetaboliteXY` if I can get closed form solutions for those...

The cardiovascular model you mentioned it could be `TwoCentral`. I don't think we can have more than one peripheral in such a model.

vjd:
This looks neat! We can/should circle back on the names ;)
Whatever the names, my suggestion is to have searchable docs for these models that will provide information on the model, the compartments and the required set of parameters. We should also have a system (much later down the line) to throw meaningful error messages when the required set of parameters are not found rather than a huge stack trace

pkofod:
A couple of points here:
 - Do we want rate formulations as well? And if so, is it okay that the interface is then `ModelNameRates` or is `ModelName{Rates}`  better? (I assume that'll work) @andreasnoack @vjd  @jony5525 
- The the three compartment is not in yet. It will be added, but I'm not sure this PR should be held back by those. We can add them (It) later. It does have metabolite models instead. Transit compartments is also 100% possible.
- More importantly: names! We shouldn't merge this with bad names, so let's find a good same of names. Just off the top of your heads, what would be your preferred names? We can discuss based on that: @vjd @andreasnoack @jony5525 @ChrisRackauckas .We have full freedom here, so let's find some good ones.

ChrisRackauckas:
>Do we want rate formulations as well? And if so, is it okay that the interface is then ModelNameRates or is ModelName{Rates} better? (I assume that'll work)

I think it would be really weird for the model to just fail if you use certain events. The overhead can't be that high?

pkofod:
> I think it would be really weird for the model to just fail if you use certain events. The overhead can't be that high?

Oh, I think we might be talking past each other. I'm not talking about a dose with a rate, I'm talking about specifying pk parameters as either "PK parameters" CL, V, ...  just being able to specify the raw rates between compartments.

ChrisRackauckas:
Oh yes, calling that the rate form makes sense.

vjd:
how about this 
I think the naming can take this form --
`<Compartment name><number of them>`
if we have multiple such compartments, chain the names together.
so, e.g.

`Central1`  - 1 compartment
`Depot1Central1`  - 1 compartment with first order absorption
`Depots2Central1` (note the plural in `Depot` ) - 1 compartment with dual absorption
`Central1Periph1` - 2 compartment
`Depot1Central1Periph1` - 2 compartment with first order absorption
`Depots2Central1Periph1`  (note the plural in Depot ) - 2 compartment with dual first order absorption
`Central1Periph1Metab1` - 1 compartment parent and 1 compartment metabolite
`Central1Periph1Metab1MetabPerip1`  2 compartment parent and 2 compartment metabolite

vjd:
All checks passed :)

src/analytical_solutions/standard_models.jl
```diff
@@ -1,59 +1,191 @@
...
+  amt₀ = amounts + doses   # initial values for cmt's + new doses
+  Λ, 𝕍 = eigen(m, p)
+
+  # We avoid the extra exp calls, but could have written:
+  # Dh  = Diagonal(@SVector(exp.(λ * (_t - _t₀)))
+  # Dp  = Diagonal(@SVector(expm1.(λ * (_t - _t₀))./λ))
+  # Instead we write:
+  Dp = Diagonal(expm1.(Λ * (t - t₀)) ./ Λ)
+  Dh = Dp .* Λ + I
+  amtₜ = 𝕍*(Dp*(𝕍\rates) + Dh*(𝕍\amt₀)) # could derive inverse here
```
ChrisRackauckas: was the speed of doing it from the eigensystem double checked?

pkofod:
> just needs a performance regression check

Yes, I will check it. If not, we can just special-case it.

pkofod:
Okay, I checked it. It was indeed slower, but Andreas and I found the source: `*(a::Diagonal(::SVector), b)`(-ish) falls back to Array operations, so it allocated. Ditched the `Diagonal` wrapper, and as Andreas witnessed, timings were the same. I merged master into it, but if tests pass this should be good to go.

pkofod:
Accidentally introduced an old name in tests with merge master.

PR #742, opened 2019-11-18
Easier optimization function changes

ChrisRackauckas:
The interface is still fully accessible, but

```julia
fit(mdsl1, data, param, Pumas.FO(), optimize_fn = DefaultOptimizeFN(verbose=true))
```

and other goodies are readily available. Is there an easy way to make changing the algorithm easier as well? The invH function makes that a little difficult.


pkofod:
`invH` is only relevant for BFGS, so we can just ignore it or error if it's set for other methods?

andreasnoack:
I think we should instead delegate the use of custom optimization routines to separate `fit` methods and document how it works. We could have a test that showed such an example. If we make that change, we could specialize the default `fit` method on Optim and the same way we specialize the ODE solves on DiffEq. It would make it easier to use features from Optim.

ChrisRackauckas:
What features would we gain over this form? It looks like this is able to pass any optimization method to any algorithm, and has all of the options of Optim available?

andreasnoack:
I don't think the current way of organizing this is robust. I'm not even sure it really works for anything but `Optim`. Does it? The current structure is pretty convoluted and it would be much more clear if we just made the default `fit` method use `Optim` directly.

I believe that `fit` methods should make use of the structure of the problem. E.g. if you want to try out a gradient free optimization method then you probably don't the current cost objective which evaluates gardients. It also might not be appropriate for some global optimization methods.

pkofod:
I think Andreas's point is that this is fine for specifying the optimizer, but once you hit `fit` you're constructing `OnceDifferentiables` that are unique to optim, etc, etc. So if you pass in an NLopt optimizer, then you need to have `OnceDifferentiable` and NLopt speak together, so this interface won't work there. That's why we might as well just bake in Optim in one dispatch branch, and tell how to create a cost function for a new package.

ChrisRackauckas:
>I think Andreas's point is that this is fine for specifying the optimizer, but once you hit fit you're constructing OnceDifferentiables that are unique to optim, etc, etc. So if you pass in an NLopt optimizer, then you need to have OnceDifferentiable and NLopt speak together, so this interface won't work there. That's why we might as well just bake in Optim in one dispatch branch, and tell how to create a cost function for a new package.

Or just make `OnceDifferentiable` have a call overload `f(p)` that uses the f-only part to produce the computed cost output? That would be nice to have just in Optim anyways, and 

https://github.com/PumasAI/Pumas.jl/blob/4652a867b621399d885c5f702ed1002852cf2297/src/estimation/likelihoods.jl#L1157-L1159

it would be efficient since the gradient calculation would just be skipped.

pkofod:
For what it's worth (nothing) that's how new optim works... but you can just say`f(p)->value(od, p)`. The point is just that the way that `fit` is written, you cannot just define a new `optimize_fn` and it'll work.

ChrisRackauckas:
It's at least an extension of what we had before that fixes the current issue, so I'm merging. Anything bigger can get a discussion, but is unrelated to this PR.

PR #746, opened 2019-11-21
Fix NCASubject(::Subject) constructor

andreasnoack:
@YingboMa I'm removing any missings from the observations during construction. I'm assuming that `NCA` requires that since the show method for `NCASubject` fails for `Union{T,Missing}` elements.


YingboMa:
I think we also need fix the show method. Could you open an issue for this?

andreasnoack:
Just filed https://github.com/PumasAI/Pumas.jl/issues/748. While preparing a minimal example I realized why the purpose of `clean` is. Wouldn't a better fix for the `NCASubject(Subject)` be to just not set `clean=false`?

YingboMa:
> Wouldn't a better fix for the NCASubject(Subject) be to just not set clean=false?

It is set to `clean=false` for performance. I think the save of computation here should be minuscule, so we perhaps should set it to `true`.

Issue #761, opened 2019-11-28
Bolus tests of proportional+additive model with FOCE sometimes causes NaNs


andreasnoack:
See e.g. https://gitlab.com/PumasAI/Pumas-jl/-/jobs/362664512. This only happens on come machine but unfortunately one of the machine is the CI machine and it results in very slow completion times for the tests. We suspect the issue is related to the use of the `DiffEqBase.fastpow`, see https://github.com/PumasAI/Pumas.jl/pull/736. Switching to e.g. `Tsit5()` instead of the default solver also seems to avoid the issue.

ChrisRackauckas:
Well this is quite interesting because it means it's not fastpow, but rather an interaction with the switching mechanism. @YingboMa that's a good hint.

andreasnoack:
This model is no longer supported so closing.

PR #762, opened 2019-11-28
Use Tsit5() solver for FOCE fit of proportional+additive model in

andreasnoack:
nlme/bolus.jl to avoid hitting issue with `NaNs` during the ODE solve.
 
This should save us quite some CI time. The underlying issue is tracked in #761 so we should revisit this some time in the future to see if it remains a problem.


andreasnoack:
This small change reduces test time for the `nlme-ml2` task from 50 minutes to 23 minutes so I'll merge to get faster CI right away.

ChrisRackauckas:
Yeah, a lot of the simple analytical tests probably don't need the autoswitch to the stiff ODE solver. However, any problem with nonlinearities will need it, so it's a good default once we fix the fastpow issue.

Issue #744, opened 2019-11-20
Broader FOCEI support


andreasnoack:
Currently, we only support the FOCEI approximation in the `Normal` and `LogNormal` cases but it shouldn't be too much work to also support other distributions. The way the code is structured, we just need to implement methods for computing the approximation of the second order derivative of the log-likelihood in the random effect, i.e. the conditional expectation of the outer product of the gradient of the log-likelihood.

Supported distributions:
- [x] `Normal`
- [x] `LogNormal`
- [ ] `Bernoulli`
- [ ] `Poisson`
- [ ] `Gamma`

Feel free to suggest other distributions.

jony5525:
In repeated time-to-event models we often use:

- [ ] Exponential
- [ ] Weibull
- [ ] Gompertz

I think all of them would be great to have.

andreasnoack:
Great I've added them to the list in the top post. Didn't know the Gompertz. It also looks like we don't have it in Distributions so it might be a little extra work to support that one.

andreasnoack:
@jony5525 Weibull and Gompertz will be used for time-to-event models where we won't have random effects, right?

jony5525:
> @jony5525 Weibull and Gompertz will be used for time-to-event models where we won't have random effects, right?

In time-to-event we usually don't have random effects (even though there are some research going on trying this) but in repeated time-to-event we always have random effects (and similar distributions as for TTE)

PR #765, opened 2019-11-29
Make FOCE estimation work for the exponential family and restrict

andreasnoack:
it to the case when the dispersion doesn't depend on random effects. This will handle most of the cases in #744. The nice thing is that it makes the calculation of the second order term simpler and for the exponential family (and also LogNormal) it ended up being sufficient with a single method. 

oS with this, you can e.g. do
```julia
julia> fit(poisson_model, df, param, Pumas.FOCE())
FittedPumasModel
Likelihood approximation:         Pumas.FOCE
Objective function value:          1904.9154
Total number of observation records:    1800
Number of active observation records:   1800
Number of subjects:                       20
------------------
         Estimate
------------------
θ₁        1.0293
θ₂        0.45185
Ω₁,₁      0.12201
------------------
```

Once I started thinking a bit about the problem here, I realized that the appropriate likelihood approximation for this would be `FOCE` and not `FOCEI`. Generally, we'd like that the dispersion parameter is free and in that case, the two methods are the same. However, if we require that `FOCE` only applies to the case where the dispersion parameter is free then the calculations suddenly go from being ugly to nice looking.


src/estimation/diagnostics.jl
```diff
@@ -166,18 +166,24 @@ function cwres(m::PumasModel,
...
-  randeffs0   = TransformVariables.transform(randeffstransform, zero(vrandeffsorth))
   randeffsEBE = TransformVariables.transform(randeffstransform, vrandeffsorth)
 
-  dist0   = _derived(m, subject, param, randeffs0)
   distEBE = _derived(m, subject, param, randeffsEBE)
 
+  _dv_keys = keys(subject.observations)
+  foreach(_dv_keys) do _key
+    if !_is_homoscedastic(distEBE[_key])
+      throw(ArgumentError("dispersion parameter is not allowed to depend on the random effects when using FOCE"))
```
ChrisRackauckas: from the tests below this seems fairly breaking. How much is gained from this specialization?

pkofod: It is meant to be breaking as far as I can see. The idea is to restrict users from doing this, because there's no reason not to use FOCEI in this case..

PR #775, opened 2019-12-02
Bump OrdinaryDiffEq to match version in next JuliaPro release

ChrisRackauckas:


andreasnoack:
Why do we need require this version? Are we using any features from 5.26?

ChrisRackauckas:
We want the versions to match for our coming demo, so we will need to make sure JuliaPro is on this release. So we might as well enforce it in Pumas.

PR #784, opened 2019-12-06
FOCE support for Categorical

andreasnoack:
I'm also bumping the Distributions requirement to the version that includes the `Categorical` that we need here.


src/estimation/likelihoods.jl
```diff
@@ -889,15 +890,41 @@ function _∂²l∂η²(obsdv::AbstractVector, dv_d::AbstractVector{<:Distributi
...
   return nl, nothing, H
 end
 
+# Categorical
+function _∂²l∂η²(obsdv::AbstractVector, dv_d::AbstractVector{<:Categorical}, ::FOCE)
+
+  nrfx = length(ForwardDiff.partials(mean(first(dv_d))))
+
+  # Initialize Hessian matrix and gradient vector
+  ## FIXME! Careful about hardcoding for Float64 here
```
ChrisRackauckas: yeah... some of these are making some Zygote stuff rough. Not a big issue right now though

PR #788, opened 2019-12-09
Add another error handling check in DosageRegimen

pkofod:
If the user doesn't specify steady state dosing, it should not be possible to set an `ii` without a number of additional doses `addl`.


pkofod:
Sorry, I see I forgot to give the PR a title >_< :)

PR #789, opened 2019-12-09
IVIVC Tests

sipah00:


sipah00:
@andreasnoack how can I see updated codecov?

andreasnoack:
Currently, it's not available from our CI because we run all the jobs in parallel and I don't think we have a way to merge the different coverage files. It might be possible to have that as a post processing job. I'll take a look. For now, you can just run `Pkg.test("Pumas", coverage=true)` on your own machine and check the `.cov` files.

andreasnoack:
So I've just enable coverage reporting from CI and updated your branch. Once tests have completed, you should be able to see the effect at https://codecov.io/gh/PumasAI/Pumas.jl

andreasnoack:
Did you notice this failure https://gitlab.com/PumasAI/Pumas-jl/-/jobs/374738482?

sipah00:
Yeah, Path to file is changed, will update in next commit

andreasnoack:
I'll merge what we have here. Then we can continue in other PRs.

PR #792, opened 2019-12-10
Submit coverage info from CI

andreasnoack:
I don't think it used to work on codecov but it looks like it's working so it should now be much easier to keep track of our coverage. While working on this, I also realized that we only tested IVIVC when running the tests sequentially. We do that before each release so nothing has regressed but this PR enables IVIVC testing when running the tests in parallel.


andreasnoack:
This only touches the CI post processing and we need it for @sipah00's PR so I'll go ahead and merge so we can see the effect on the IVIVC PR.

Issue #754, opened 2019-11-22
Way to construct Subject from simulations


andreasnoack:
We should have something like a `Subject(::SimulatedObservations)` constructor such that we can easily use simulated observations for fitting. Right now we "export" and "import" via a DataFrame which is inconvenient and slow.

I think we can probably just copy the wrapped subject of the `SimulatedObservations` and then populate it the right observations and times. We can use the `dv` names in the wrapped `Subject` to figure out what is a dependent variable among all the possible simulation output.

A final comment. I my opinion, it would be simpler and more modular if we followed standard statistical terminology and had something like
- `simobs` just simulating observations, i.e. the dependent variable of a model  
- a `predict` method for computing the predictions (conditional means)
- "derived" quantities computed as a separate step after the model step based on the output from `simobs` or `predict`.

vjd:
I think this is long overdue and a good suggestion. 

>I think we can probably just copy the wrapped subject of the SimulatedObservations and then populate it the right observations and times.

some sort of `join` ? as the wrapped subject may only have events and demographics

>simobs just simulating observations, i.e. the dependent variable of a model

This is what it does now, correct?

>a predict method for computing the predictions (conditional means)

a `predict` method that takes a model, data and set of parameters instead of a `fpm`?

>"derived" quantities computed as a separate step after the model step based on the output from `simobs` or `predict`.

you mean for e.g. NCA steps? I am not sure I understand your last point

ChrisRackauckas:
>"derived" quantities computed as a separate step after the model step based on the output from simobs or predict.

That's the difference between `solve` and `simobs`. `derived` is then the middle ground that gives out the distributions without applying the `observed` block. There is a missing idea there for `predict` which pushes forward the means of the distributions.

andreasnoack:
> This is what it does now, correct?

I wouldn't say so. Currently, `simobs` will dump everything from the `@derived` block in the simulation output. The way we set up a model in Pumas is in a sense an incomplete specification of a statistical model. It's only in the `Subject` that we specify what is the dependent variable and the statistical model will change depending on which variables have been included in the `observations` part of of the `Subject`.

> a predict method that takes a model, data and set of parameters instead of a fpm?

It should still allow for passing a `FittedPumasModel`. The idea was mainly to have separate functions for computing the predictions and the simulations and observations.

> you mean for e.g. NCA steps? I am not sure I understand your last point

[...]

PR #796, opened 2019-12-10
FOCE support for negative binomial

andreasnoack:
See the added test for an example.


test/nlme/negativebinomial.jl
```diff
@@ -0,0 +1,57 @@
...
+
+  param = init_param(negativebinomial_model)
+  randeffs = init_randeffs(negativebinomial_model, param)
+
+  @test solve(negativebinomial_model, df[1], param, randeffs) isa Pumas.NullDESolution
+  @test simobs(negativebinomial_model, df, param, randeffs) != nothing
+
+  res = simobs(negativebinomial_model, df, param, randeffs)
+
+  for _approx in (Pumas.FOCE(), Pumas.LaplaceI())
```
pkofod: Should we test `vcov` explicitly here? Just to be sure it goes through.

pkofod:
I see the Subject(Simulated...) snuck its way into this PR :) Great. LGTM

andreasnoack:
> I see the Subject(Simulated...) snuck its way into this PR

Yes. I had to use simulated values instead of the Poisson data since the corresponds to an infinite dispersion parameter and that makes covariance matrix indefinite. Hence, I decided to add the `Subject` constructor to make it easier to use the simulated values so this actually also fixes #754.

PR #797, opened 2019-12-10
Deprecate Laplace in favor of LaplaceI since there they are identical

andreasnoack:
when the dispersion parameter doesn't depend on data and generally
misspecified when it does depend on the data. Furthermore, because we rely in AD, there is no
computational benefit from the current Laplace.

Once we have had this deprecation in for a while, I think we should consider renaming `LaplaceI` to just `Laplace`, but that is a separate issue.


pkofod:
> Once we have had this deprecation in for a while, I think we should consider renaming `LaplaceI` to just `Laplace`, but that is a separate issue.

Are you sure about that? :O Isn't the Laplace name the only reason why we cannot export the likelihood approximation names? :) I'll review tomorrow

pkofod:
I agree that this is the right decision. It mirrors recent FOCE changes - changes that are for the better.

Issue #734, opened 2019-11-13
Dual number issue when using autodiff in ODE solver with FO


ChrisRackauckas:
MWE:

```julia
using Pumas

theopmodel_solver_fo = @model begin
  @param begin
    θ₁ ∈ RealDomain(lower=0.1,    upper=5.0, init=2.77)
    θ₂ ∈ RealDomain(lower=0.0008, upper=0.5, init=0.0781)
    θ₃ ∈ RealDomain(lower=0.004,  upper=0.9, init=0.0363)
    θ₄ ∈ RealDomain(lower=0.1,    upper=5.0, init=1.5)
    Ω ∈ PSDDomain(3)
    σ_add ∈ RealDomain(lower=0.001, init=0.388)
    #σ_prop ∈ RealDomain(init=0.3)
  end

  @random begin
    η ~ MvNormal(Ω)
  end

  @pre begin
    Ka = SEX == 0 ? θ₁ + η[1] : θ₄ + η[1]
    K  = θ₂+ η[2]
    CL = θ₃*WT + η[3]
    V  = CL/K
    SC = CL/K/WT
  end

  @covariates SEX WT

  @vars begin
    conc = Central / SC
    cp   = Central/V
  end

  @dynamics begin
      Depot'   = -Ka*Depot
      Central' =  Ka*Depot - (CL/V)*Central
  end

  @derived begin
    dv ~ @. Normal(conc,sqrt(σ_add))
  end
end

theopp = read_pumas(example_data("event_data/THEOPP"),cvs = [:SEX,:WT])

param = (
  θ₁ = 2.77,   #Ka MEAN ABSORPTION RATE CONSTANT for SEX = 1(1/HR)
  θ₂ = 0.0781, #K MEAN ELIMINATION RATE CONSTANT (1/HR)
  θ₃ = 0.0363, #SLP  SLOPE OF CLEARANCE VS WEIGHT RELATIONSHIP (LITERS/HR/KG)
  θ₄ = 1.5,    #Ka MEAN ABSORPTION RATE CONSTANT for SEX=0 (1/HR)
  Ω  = diagm(0 => [5.55, 0.0024, 0.515]), # update to block diagonal
  σ_add = 0.388
  # σ_prop = 0.3
     )

o = fit(theopmodel_solver_fo, theopp, param, Pumas.FO(), alg=Rosenbrock23())
```

[...]

PR #805, opened 2019-12-25
Jacobian parameter calling and fixing of col type

### LINKED ISSUE #830: Pre should be initialized by time values that are matching in type.
Straightforward from the title. Introduced in https://github.com/PumasAI/Pumas.jl/pull/824 . This `col(0)` (`pre(0)`) call at the start may introduce type-instabilities or cause error if dual numbers and if time is directly used.

ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/734 for all practical purposes, because now the Jacobian code won't be hit at all. Fixes #830 at the same time.


ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/830

ChrisRackauckas:
Turns out this test was finding some deeper Jacobian things we should work on. Currently hitting an MTK issue.

ChrisRackauckas:
Waiting on https://github.com/JuliaDiffEq/ModelingToolkit.jl/pull/241

PR #806, opened 2019-12-25
Switch to all AD instead of mixed FD and AD for compution gradient of the marginal likelihood function.

andreasnoack:
Patrick and I spent some time looking into why we had problems with estimating the model with time varying covariates that is added as a test in this PR. We realized that for that model, as well as a few other models with time varying covariates, the gradients were very noisy. If we reduced the ODE solver tolerances, the optimization went better but at the cost of more time spent on solving the ODEs.

There seemed to be two major reasons why the gradients were so noisy. 1) the finite difference derivative step size had been tuned for the models with linear ODEs but the error in the solution is ~100x larger for non-linear ODEs so the step sizes where too small. 2) the error was amplified by the use of Almquist's fast method for computing the gradient, i.e. the naive finite difference gradient with recomputation of the EBEs was much more robust to noise but also much costlier because of the extra EBE calculations.

After some experimentation and discussion, Patrick and I came to the conclusion that our best option was to switch to automatic differentiation for the partial derivates of the marginal likelihood in `θ`. This makes the gradient much more precise at the cost of extra compile time. However, the extra compile time isn't too bad and the change will 1) make it possible to estimate models with time varying covariates which currently isn't practical because of noisy gradient, 2) make each iteration much faster so for non-toy models switching to a pure AD seems to be the dominating strategy.

I've added the new test with time varying covariates as a separate test group since it's still super slow. Once we have a more efficient handling of time varying covariates (which is getting closer now), the tvcov test should become much faster.


ChrisRackauckas:
The main reason for the finite difference at first was the massive compile times in v0.6 when the partials got too large. It was like 10 minutes or so at one point early on. But not that seems to be gone (IIRC the tuple length change is what we tracked it to), so I don't see a reason to not just nest dual numbers. We still can't nest duals in the nonlinear solver itself, but that is probably unnecessary.

src/estimation/likelihoods.jl
```diff
@@ -531,62 +531,42 @@ function marginal_nll_gradient!(g::AbstractVector,
                                 vrandeffsorth::AbstractVector,
                                 approx::Union{FOCE,FOCEI,LaplaceI},
                                 trf::TransformVariables.TransformTuple,
-                                args...;
-                                # We explicitly use reltol to compute the right step size for finite difference based gradient
-                                reltol=DEFAULT_ESTIMATION_RELTOL,
```
ChrisRackauckas: keep the default reltol drop?

ChrisRackauckas: Nevermind, it's on `_derived`

pkofod:
I guess we do need to monitor the compilation times somehow, but hopefully it'll only get better. Regressions have happened in the past though.

Anyway, the sheer precision improvement here is too hard to ignore, so it's the right thing to do. We can revisit if we see any disadvantages, or if some models can do without, but from what Andreas and I observed, this change was crucial for convergence in tvcov models.

PR #807, opened 2019-12-27
Update GSA to new interface and use dataframes to store results.

Vaibhavdixit02:
This PR updates the Pumas GSA wrapper to match updates in DiffEqSensitivity and also switches to using DataFrame to store the result. Right now the tests are failing as this is waiting on a new release of DiffEqSensitivity.


ChrisRackauckas:
@Vaibhavdixit02 it would be good to finish this one up now that the tags are in.

PR #815, opened 2020-01-14
Reduce chunk size in AD to make compilation faster

andreasnoack:
Also, fix computation of marginal likelihood for multiple dvs and add a test for it.

The HCV model revealed that using the default chunk sizes in ForwardDiff caused extreme compile times. The solution is to bound the chunk sizes for Jacobians and Hessians. This significaly improves out compile times at the cost of more calls to the objective function but notice that we'll still have fewer calls than we'd have with noisy finite difference derivatives.

Since our tests are compile heavy, it looks like this change makes the ML1 tests 10-15 minutes faster than they currently are.


vjd:
LGTM as all checks have passed

PR #824, opened 2020-01-28
Make pre a function of time to facilitate efficient tvcov handling

pkofod:
This PR is almost finished, but there are some regressions that I'm still looking at. I will write a more details explanation later today. You can read the code, but I will have to change some things, so maybe don't give the extensive review just yet :)

**EDIT**
Right, so let's get into the details.

This PR has a few significant, and breaking, changes so it's important to stay awake. Would have wanted this PR in sooner, but various technical things meant that I had to rework stuff several times.

**No more `@tvcov`**
Previously, this would be needed to do time-varying covariates.
```julia
    @pre begin
      _pnad = @tvcov pnad time
      _gawk = @tvcov gawk time
      _wtkg = @tvcov wtkg time
      mat = t -> _pnad(t) / (ec50 + _pnad(t))
      cl = t -> tvcl * mat(t) * (_gawk(t)/34)^gaeffect * (1 - mat(t)) * _wtkg(t)^0.75 * exp(η[1])
      v = t -> tvv * _wtkg(t) * exp(η[2])
      ka = tvka
    end
```
This was bad for several reasons. The whole explanation of macros and `t->` function definitions were "too special" for time varying covariates compared to constant covariates. The new `@pre` block now just uses the variable names:
```julia
    @pre begin
      mat = pnad / (ec50 + pnad)
      v = tvv * wtkg * exp(η[2])
      cl = tvcl * mat * (gawk/34)^gaeffect * (1 - mat) * wtkg^0.75 * exp(η[1])
      ka = tvka
    end
```
**Implicit time references**
Before, the dynamics block was partially implicit in its time references:
```julia
@dynamics begin
  X' = -k*X
```
So there is no explicit mention of `(t)` on the dynamic variables. However if `k` was time varying, you would write
```julia
@dynamics begin
  X' = -k(t)*X
end
```
So it was partially implicit because `X, X'` have no time references, but `k(t)` does. Now, time is completely implicit, which I think is great for users.

**New job for `read_pumas` and the `Subject` constructor**
All of this works by having the `Subject` constructor handle all the tvcov-work. In this PR there are no bells and whistles in the sense that I will have to come back and add more features in terms of other interpolation styles, and also actually use DataInterpolations directly.

The `Subject` constructor now takes all covariates, parses them, looks for time-varying covariates, constant covariates, and no covariates, and turns them into appropriate interpolants. Then, `@pre` automatically evaluates them at `t`, such that they can just be references by their covariate names, no need to explicitly reference `t`.

**`param`, `randeffs`, and `@pre`**
`param` and `randeffs` are *no longer* packaged with the actual lhs `@pre`-variables! This means that to use parameters directly in `@dynamics`, they have to a left-hand side in `@pre`.  This has to do with the way `@derived` is defined. I wanted to make `@derived` more like `@pre` and `@dynamics` in terms of time referencing, but that was impossible because we need `@nca` support in `@derived`. It's a small price to pay, and it is easy to explain: need a var in `@dynamics`? Define it in `@pre`!

So before you could do
```julia
...
@param begin
  theta = ...
end
@dynamics begin
  X' = -theta[1]*X
end
```
now you have to do
```julia
@param begin
  theta = ...
end
@pre begin
  ka= theta[1]
end
@dynamics begin
  X' = -ka*X
end
```
(or you could of course also just make the full vector available). I think it's a small price to pay, and it actually makes the returned namedtuple from pre smaller, which means fewer problems with tuples sizes in numtype!

**d_discontinuities**
So, an additional discussion point has been `d_discontinuities`. I did not see significant speed improvements, but I did see some behavioral differences, so I've turned it off for now until we can investigate in the future.


src/simulate_methods/diffeqs.jl
```diff
@@ -21,16 +22,18 @@ function _build_diffeq_problem(m::PumasModel, subject::Subject, args...;
...
   Tt = promote_type(numtype(tstops), numtype(tspan))
   tspan = Tt.(tspan)
   :callback ∈ keys(prob.kwargs) && (cb = CallbackSet(cb, prob.kwargs[:callback]))
-
   # Remake problem of correct type
   new_f = make_function(prob,fd)
-
   remake(m.prob; f=new_f, u0=Tu0, tspan=tspan, callback=cb, saveat=saveat,
-                 tstops = tstops,
+                 tstops = tstops,# d_discontinuities=d_discontinuities,
```
andreasnoack: Is this supposed to be commented out?

src/data_parsing/data_types.jl
```diff
@@ -198,16 +198,18 @@ Fields:
...
-struct Subject{T1,T2,T3,T4}
+struct Subject{T1,T2,T3,T4,T5,T6}
   id::String
   observations::T1
   covariates::T2
   events::T3
   time::T4
-
-  function Subject(data, Names,
+  tvcov::T5
```
andreasnoack: Maybe merge these last two fields into the current `covariates` and add generic function for extracting the covraite time.

src/simulate_methods/analytical.jl
```diff
@@ -21,7 +21,7 @@ function _build_analytical_problem(m::PumasModel, subject::Subject, tspan, col,
                                    args...; kwargs...)
   f = m.prob isa ExplicitModel ? m.prob : m.prob.pkprob
   u0 = pk_init(f)
-  numtypecol = numtype(col)
+  numtypecol = col isa Function ? numtype(col(0)) : numtype(col)
```
andreasnoack:
```suggestion
  numtypecol = numtype(col(0))
```

src/simulate_methods/diffeqs.jl
```diff
@@ -299,7 +301,7 @@ function ith_subject_cb(p,datai::Subject,u0,t0,ProbType,saveat,save_discont,cont
     end
   end
   save_positions = save_discont ? (true, true) : (false, false)
-  tstops,DiscreteCallback(condition,affect!,subject_cb_initialize!,save_positions)
+  tstops,DiscreteCallback(condition,affect!,subject_cb_initialize!,save_positions)#,d_discontinuities
```
andreasnoack:
```suggestion
  tstops,DiscreteCallback(condition,affect!,subject_cb_initialize!,save_positions)
```

test/core/error_handling.jl
```diff
@@ -89,7 +89,8 @@ param = (θ = θ,
     Σ_dv = 0.04,
     Σ_pddv = 1)
 
-@test_throws MethodError conditional_nll(model, population[1], param, (η=zeros(9),))
+
+#@test_throws MethodError conditional_nll(model, population[1], param, (η=zeros(9),))
```
andreasnoack: Is this supposed to be commented out?

test/core/multiresponses.jl
```diff
@@ -67,7 +68,7 @@ m23 = @model begin
     @derived begin
         # TODO: allow direct output of dynamical variables
         ev1    = Ev1
-        cp     = Cent / θ[3]
+        cp     = Cent ./ θ[3]
```
andreasnoack:
```suggestion
        cp     = Cent / θ[3]
```

test/core/static_array_test.jl
```diff
@@ -101,18 +104,23 @@ p = ParamSet((θ = VectorDomain(3, lower=zeros(3), init=ones(3)),
...
 function post_f(col,u,t)
-    (conc = u[2] / col.V,)
+   col_t = col(0.0) # no t needed as covariates are constant
+    (conc = u[2] / col_t.V,)
 end
-function derived_f(col,sol,obstimes,subject)
-    central = sol(obstimes;idxs=2)
-    conc = @. central / col.V
-    (conc = conc,)
+function derived_f(col,sol,obstimes,subject, param, randeffs)
```
andreasnoack:
```suggestion
function derived_f(col, sol, obstimes, subject, param, randeffs)
```

test/core/time_varying_covar.jl
```diff
@@ -22,11 +22,14 @@ function rfx_f(p)
...
-    CL = param.θ[2] * ((cov.wt/70)^0.75) *
-         (param.θ[4]^cov.sex) * exp(randeffs.η[1]),
-    V  = param.θ[3] * exp(randeffs.η[2]))
+  function __pre(t)
+    cov = subject.tvcov(t)
+    Ka = t*param.θ[1]  # pre
+    CL = param.θ[2] * ((cov.wt/70)^0.75) * (param.θ[4]^cov.sex) * exp(randeffs.η[1])
+    V  = param.θ[3] * exp(randeffs.η[2])
+
+    return (CL=CL, V=V, Ka=Ka, param..., randeffs...)
```
andreasnoack:
```suggestion
    return (CL=CL, V=V, Ka=Ka)
```

test/features/discrete_stochastic.jl
```diff
@@ -37,11 +39,14 @@ jump_prob = JumpProblem(prob,Direct(),jump1,jump2)
 
 init_f = (col,t) -> [0.0,0.0]
 
-function derived_f(col,sol,obstimes,subject)
-    central = sol(obstimes;idxs=2)
-    conc = @. central / col.V
-    dv = @. Normal(conc, conc*col.Σ)
-    (dv=dv,)
+function derived_f(col,sol,obstimes,subject, param, randeffs)
```
andreasnoack:
```suggestion
function derived_f(col, sol, obstimes, subject, param, randeffs)
```

test/nlme/poisson_model.jl
```diff
@@ -19,12 +19,13 @@ using Pumas, Test
...
       baseline = θ₁*exp(η[1])
       d50 = θ₂
+      dose_d50 = dose/(dose + d50)
     end
 
     @covariates dose
 
     @derived begin
-      dv ~ @. Poisson(baseline*(1-dose/(dose + d50)))
+      dv ~ @. Poisson(baseline*(1-dose_d50))
```
andreasnoack:
```suggestion
      dv ~ @. Poisson(baseline*(1 - dose_d50))
```

test/nlme/tvcov.jl
```diff
@@ -133,76 +128,76 @@ df = identity.(df)
...
-# tvv           3.7096
-# tvka          0.88369
-# ec50         15.0
-# gaeffect     10.97
-# Ω₁,₁          0.071157
-# Ω₂,₂          0.085358
-# σ_prop        0.040622
-# -----------------------
-# """
+        tvcov_model_normal,
```
andreasnoack: Identation

src/simulate_methods/diffeqs.jl
pkofod: Yes, because adding them in gives odd behavior (lower likelihoods, etc...) see the discussion in the first post.

andreasnoack:
Since there haven't been any new comments, I'll go ahead and merge this such that we can start using it and figure out which adjustments are required.

ChrisRackauckas:
This is a pretty big change to merge right before the comments...

ChrisRackauckas:
d_discontinuities
So, an additional discussion point has been d_discontinuities. I did not see significant speed improvements, but I did see some behavioral differences, so I've turned it off for now until we can investigate in the future.


It should have a behavioral difference because it'll be more correct. That's a reason to keep it on.

src/simulate_methods/analytical.jl
```diff
@@ -21,7 +21,7 @@ function _build_analytical_problem(m::PumasModel, subject::Subject, tspan, col,
                                    args...; kwargs...)
   f = m.prob isa ExplicitModel ? m.prob : m.prob.pkprob
   u0 = pk_init(f)
-  numtypecol = numtype(col)
+  numtypecol = numtype(col(0))
```
ChrisRackauckas: Wrong type on the input

src/simulate_methods/diffeqs.jl
```diff
@@ -5,9 +5,10 @@ function _build_diffeq_problem(m::PumasModel, subject::Subject, args...;
   prob = typeof(m.prob) <: DiffEqBase.AbstractJumpProblem ? m.prob.prob : m.prob
   tspan = prob.tspan
   col = prob.p
+  col0 = col(0) # would be great to get the numtype without this
   u0 = prob.u0
 
-  T = promote_type(numtype(col), numtype(u0), numtype(tspan))
+  T = promote_type(numtype(col0), numtype(u0), numtype(tspan))
```
ChrisRackauckas: Should match time type

src/simulate_methods/diffeqs.jl
```diff
@@ -21,16 +22,18 @@ function _build_diffeq_problem(m::PumasModel, subject::Subject, args...;
   ft = DiffEqBase.parameterless_type(typeof(prob.f))
 
   # figure out callbacks and convert type for tspan if necessary
+  # d_discontinuities are used to inform diffeq about the places where things change
+  # suddenly in the model and introduce discontinuities in the derivates (such as
+  # time varying covariates etc)
   tstops,cb = ith_subject_cb(col,subject,Tu0,tspan[1],typeof(prob),saveat,save_discont,continuity)
+  # tstops,cb,d_discontinuities = ith_subject_cb(col,subject,Tu0,tspan[1],typeof(prob),saveat,save_discont,continuity)
```
ChrisRackauckas: Why renove,?

test/core/error_handling.jl
ChrisRackauckas: Indeed this looks ofd

PR #825, opened 2020-01-29
Add StructArrays to project.toml to prepare for tvcov

pkofod:
Edit: Editting to reflect that it is indeed to start the testing for the next release, as mentioned below.


andreasnoack:
This is useful because we can then start the release process of Pumas while we are waiting for on #824.

PR #828, opened 2020-02-06
Enable symbolic Jacobian and W factorizations

ChrisRackauckas:
Enables ModelingToolkit to use the generated Jacobian and W-fact function for precomputed symbolic factorizations of I-gamma*J to make the implicit ODE solvers really fly. Usually this gives around a 2x speedup, so we'll see what we get here. The factorization can take awhile to compute on really large systems, so I capped it at 16 at which it no longer applies this optimization.


src/dsl/model_macro.jl
```diff
@@ -339,11 +342,20 @@ function dynamics_obj(odeexpr::Expr, pre, odevars, callvars, bvars, eqs, isstati
     rhseq = eq.args[3]
     push!(mteqs,lhsvar ~ convert_rhs_to_Expression(rhseq,bvars,dvars,params,t))
   end
-  f_ex = generate_function(ODESystem(mteqs),dvars,params)[1]
+  f_ex = ModelingToolkit.generate_function(ODESystem(mteqs),dvars,params)[1]
+  J_ex = ModelingToolkit.generate_jacobian(ODESystem(mteqs),dvars,params)[1]
```
YingboMa: Do we want to have a cap for the Jacobian, too?

ChrisRackauckas: I don't think so. Even the crazy QSP models have had fine Jacobian calculations. The 25,000 term BCR does fine with Jacobians. So, it seems like it's fine.

ChrisRackauckas: It we run into a user issue we can backtrack it, but I would be heavily surprised if a Pumas model pushes it that far.

ChrisRackauckas: factorization is a whole different story though.

PR #829, opened 2020-02-06
re-enable d_discontinuities

ChrisRackauckas:
This tells the ODE solver to re-eval FSAL at the points of derivative discontinuities which is required for maintaining higher order and should result in fewer time steps and faster solves. @pkofod could you explain exactly what you saw when testing before?


pkofod:
Just different objective values upon convergence than the reference ones in the test. Without d_discontinuities we matched nonmem (Andreas ran it for me) and with we had a different value.

ChrisRackauckas:
Different, as in worse?

pkofod:
> Different, as in worse?

Yes, larger objective value, but I think it was expected come to think about it. I believe that the data set we estimated was simulated from a solution without d_discontinuities. In that case, then the method that solves it "wrong" will be able to fit a "better model" in the sense that the data generating process and the estimation procedure are in sync (at the same parameter values they will predict the same behavior at the end-points). 

So in principle, I'm all for enabling it, I just hadn't done any investigation into why there was a difference there, so I figured the default was fine.

If we make this change (I if it's more precise then why not?) we just have to be aware that we should expect our solutions, estimates, and optimal objective function values to generally be different from nonmem. I mean, it's not a good argument to be imprecise because nonmem is, we just need to maybe write it somewhere and be able to explain it to users who will want to compare.

ChrisRackauckas:
Needs https://github.com/JuliaDiffEq/DelayDiffEq.jl/pull/172

ChrisRackauckas:
>Yes, larger objective value, but I think it was expected come to think about it. I believe that the data set we estimated was simulated from a solution without d_discontinuities. In that case, then the method that solves it "wrong" will be able to fit a "better model" in the sense that the data generating process and the estimation procedure are in sync (at the same parameter values they will predict the same behavior at the end-points).

How much of a difference is it making? How is the data being generated? Indeed it should just get a more precise solve, which is why I'm a bit confused. NONMEM with really really low tolerances should give similar values though, just using time stepping adaptivity to overcome the discontinuity. 

Hairer demonstrates this quite nicely:

Capture

ChrisRackauckas:
Wow, Hairer wasn't kidding that it's much less dramatic than for multistep codes. Here's what NONMEM should be doing if it's doing it right without declaring the discontinuities:

https://hal.archives-ouvertes.fr/hal-01928590/document

Capture

They have to either be declaring discontinuities or they are cutting corners.

pkofod:
> How much of a difference is it making? How is the data being generated?

Indeed, as we discussed on slack the real benchmark is a super accurately simulated data set.

ChrisRackauckas:
The DelayDiffEq issue has been handled, so now it's just a choice @pkofod .

pkofod:
I trust you :) If the only reason why we get a different result is that we're more precise I say we go for it .

pkofod:
I think you just have to change the reference likelihood value here.

Issue #787, opened 2019-12-08
default `bioav` to 1 when explicitly not set


vjd:
This model below produces a bounds error 

```
model = @model begin
    @param begin
        tvcl ∈ RealDomain(lower=0, init = 1.0)
        tvv ∈ RealDomain(lower=0, init = 100)
        tvka ∈ RealDomain(lower = 0, init= 0.5)
        tvlag ∈ RealDomain(lower = 0, init = 0.5)
        tvbio ∈ RealDomain(lower = -12, upper = 12)
        Ω ∈ PDiagDomain(init=[0.04, 0.04])
        σ ∈ RealDomain(lower=0,init=0.01)
    end
    @random begin
        η ~ MvNormal(Ω)
    end
    @pre begin
      CL = tvcl *  exp(η[1])
      V  = tvv
      Ka = tvka
      lags  = (Depot = tvlag,)
      flgt = tvbio + η[2]
      f1 = exp(flgt)/(1 + exp(flgt))
      bioav = (Depot = f1,)
    end
    @dynamics Depots1Central1

       @derived begin
           cp = Central/V
           dv ~ @. Normal(cp, abs(cp)*σ)
         end
end
biolag_params = (tvcl = 1.0,
                tvv = 100.0,
                tvka = 0.5,
                tvlag = 0.5,
                tvbio = 0.5,
                Ω = Diagonal([0.05, 0.05]),
                σ = 0.02)
s1 = Subject(id=1, evs=DosageRegimen(100, cmt=1, time=0))
s2 = Subject(id=2, evs=DosageRegimen(100, cmt=2, time=0))
pop = Population([s1,s2])
sims = simobs(model, pop, biolag_params,obstimes = [0,2.5,5,10,15,30])
```
It errored  on the second subject because `bioav` into central doesn't have a value. we'd need to add that if the keycheck find the index does not exist, then default it to 1. Currently the default is set to 1 only when the `bioav` is not specified at all.

This model can be fixed by specifying ` bioav = (Depot = f1, Central =1)`

Issue #649, opened 2019-08-28
estimate dose control parameters


vjd:
I will post a reproducible example soon once some of the PR's are merged

andreasnoack:
On master, I'm getting
````
julia> o = fit(iv,data,param,Pumas.FOCEI())
FittedPumasModel

Successful minimization:                true

Likelihood approximation:        Pumas.FOCEI
Objective function value:            14.3725
Total number of observation records:     288
Number of active observation records:    288
Number of subjects:                       24

----------------------
           Estimate
----------------------
tvcl       11.065
tvv        49.12
dur         1.0032
Ω₁,₁        0.046427
Ω₂,₂        0.056934
σ_prop      0.0010331
----------------------
```
so has this been fixed?

pkofod:
Any chance it's because something was incorrectly handled in the dataframe construction, that was fixed here ? https://github.com/PumasAI/Pumas.jl/pull/489

ChrisRackauckas:
Quite possibly. We should still get this added as a test.

andreasnoack:
> Any chance it's because something was incorrectly handled in the dataframe construction, that was fixed here ? #489

Doesn't seem to be it. I can estimate `dur` even in Pumas 0.2. The variance is just tiny so sometimes the show method rounds to `1.0`.

ChrisRackauckas:
Try fitting with a different initial from the true value? I would assume the gradient is still weird due to the lack of time duals.

andreasnoack:
We tried starting it in 3.0 and it still found the ~1.0 solution.

ChrisRackauckas:
Try having a bioav as well? We should make a test that just really loads up on the DCPs.

ChrisRackauckas:
This could be another example of an AD bug accidentally fixed by https://github.com/JuliaDiffEq/OrdinaryDiffEq.jl/pull/841

andreasnoack:
Same with OrdinaryDiffEq@5.12.0. Notice that it's not clear from the original post that there is an issue.

ChrisRackauckas:
indeed. Try other dcps

pkofod:
> indeed. Try other dcps

[...]

Issue #538, opened 2019-06-12
time varying covariates on DCP's


vjd:
I think the problem that this model is not providing NCA outputs using `@derived` is because of the presence of time-varying covariates and the fact that time is a covariate. I may be wrong. Note that you need the master Bioequivalence.jl for the `generate_designs` function

```
using PuMaS
using Bioequivalence
using LinearAlgebra

bemodel = @model begin
    @param begin
        TVCL = RealDomain(init=10)
        TVV = RealDomain(init=500)
        TVKA = RealDomain(init=1)
        TVF = RealDomain(init=1)
        Ωbsv = VectorDomain(4)
        Ωwsvr = VectorDomain(1)
        Ωwsvt = VectorDomain(1)
        Ωsbf = VectorDomain(1)
    end

    @random begin
        η ~ MvNormal(Ωbsv)
        κ ~ MvNormal(Ωwsvr)
        γ ~ MvNormal(Ωwsvt)
        α ~ MvNormal(Ωsbf)
    end

    @pre begin
        CLi = TVCL*exp(η[1])
        Vi = TVV*exp(η[2])
        Kai = TVKA*exp(η[3])
        biaov = if formulation == "tablet"
            [1,TVF*exp(η[4] + κ[1])]
        else
            [1,TVF*exp(gmr)*exp(η[4] + γ[1] + α[1])]
        end # if

    end

    @covariates sequence period formulation gmr

    @dynamics begin
        Gut' = -Kai*Gut
        Central' = Kai*Gut -(CLi/Vi)*Central
    end

    @derived begin
        cp = (Central / Vi)
        nca := @nca cp
        #auc =  NCA.auc(nca)
        #cmin = NCA.cmin(nca)
    end

end

p = (TVCL = 10,TVV = 500,TVKA = 1,TVF = 1,
Ωbsv = diagm(0 => [0.04,0.04, 0.04,0.04]),
Ωwsvr = diagm(0 => [0.04]),
Ωwsvt = diagm(0 => [0.04]),
Ωsbf = diagm(0 => [0.04]),
)

des = generate_design("2S4P2", [1000,1000], ["tablet", "capsule"], [12,12])
des[:dv] = 0
des[:gmr] = -0.223144
des[:ss] = 1

df = process_nmtran(des, [:sequence, :period, :formulation, :gmr])
[...]

PR #833, opened 2020-02-08
Event-based DCP Overhaul

ChrisRackauckas:
Makes the DCP calculations directly based on events at their times. Fixes https://github.com/PumasAI/Pumas.jl/issues/538 , Fixes https://github.com/PumasAI/Pumas.jl/issues/787 . Fixes https://github.com/PumasAI/Pumas.jl/issues/649 for duration, rate, and bioav (except when SS). Shows that fitting lags by direct AD isn't possible in our setup (we can work around it, it's not easy, leaving for later).


ChrisRackauckas:
The weird thing is what to do with time-based is that, what is time? So the lag is calculated at `pre(ev.time)`, and then the other DCPs are at `pre(ev.time + lag)`, since that's the actual time of the dosing, if `lag` exists in there. It's a little odd but it's the most correct model I could think of.

ChrisRackauckas:
So it looks like the complication in the DCP fitting should only exist when there are steady states, except for lags.

ChrisRackauckas:
I'm only testing the AD here, so someone else should do a follow-up PR to add tests for the fitting. But if the AD works that should be fairly straightforward.

ChrisRackauckas:
AD cannot get derivatives of lags. It's the same problem as https://discourse.julialang.org/t/differentiating-through-a-jump-problem/31496/11?u=chrisrackauckas . Because the discontinuity is modeled discretely, there is no connection between the dose time and the `u` which gives a direct effect from the duals. This could be handled more manually in our case, but that's a whole ballgame I'm going to have to leave for later.

ChrisRackauckas:
Issues to open after this PR:

- [ ] Tests fitting rate and duration. This only shows AD should always work, which tells us fitting should work, but doesn't directly test fitting.
- [ ] Tests fitting bioav. Currently will fail on steady state dosing.
- [ ] Issue on fitting with ContinuousCallbacks. Need to handle the t numtype in that case.
- [ ] Issue on fitting bioav with steady state dosing
- [ ] Issue on fitting lags

We can close the old DCP fitting issue though since its examples should all be handled.

ChrisRackauckas:
I ended up handling the ContinuousCallbacks AD issue, so now it just needs tests, so that bullet was changed.

pkofod:
This looks good. Conceptually, lags is a bit strange in my view, but changes look good.

ChrisRackauckas:
Yeah, I'm not sure time-varying lags actually come up... but if they do, at least it has a sensible interpretation.

PR #834, opened 2020-02-11
NCA: handle route being the floating point 'Inf'

YingboMa:
`"inf"` in a CSV file can be parsed as a floating point number `Inf`.


ChrisRackauckas:
🤣 that's a fun bug.

PR #841, opened 2020-02-11
Fix some cases where args... and kwargs... weren't correctly passed

andreasnoack:
Simplify some of the dispatch in likelihoods.jl and diagnostics.jl.
Part of this is the removal of the remaining parts of the NONMEM-like
API for predictions and residuals.

Most of these methods weren't user-facing in practice anyway since they didn't have
methods for `FittedPumasModel` so the changes are mostly internal and
greatly simplifies the code.

This PR also fixes an issue that caused the EBEs to be reestimated when showing fitted models instead of just reusing the saved EBEs. For big models, it caused the show methods to hang while the EBEs were estimated.


src/estimation/likelihoods.jl
```diff
@@ -493,6 +432,81 @@ function marginal_nll(m::PumasModel,
...
+  # early in case the parameter values cause the likelihood to be Inf. This
+  # can e.g. happen if the ODE solver can't solve the ODE for the chosen
+  # parameter values.
+  if isinf(nll1)
+    return nll1
+  end
+
+  # The different parallel computations are separated out into functions
+  # to make it easier to infer the return types
+  if parallel_type === Serial
```
ChrisRackauckas: We should just make this be `ensemblealg` and use the same ensemble algorithms, even if computed differently, just for interface consistency.

ChrisRackauckas: I'll do this in a follow-up

andreasnoack: We can due that but notice that I'm just reordering some methods here. This method is not being changed in this PR so it's indeed better to handle in a separate PR.

ChrisRackauckas: Yeah it just came to mind because it showed up.

Issue #844, opened 2020-02-11
eiwres should be parallelized


ChrisRackauckas:
It's the one diagonistic not handled in https://github.com/PumasAI/Pumas.jl/pull/843 . The right way to do this would be to make a dispatch for `_derived` on populations that parallelizes

PR #845, opened 2020-02-11
fix and parallelize eiwres

ChrisRackauckas:
Implements a hook for _derived to call simobs on a population and just get the derived appropriately calculated and returned, and then uses this population-based _derived in eiwres. However, I also noticed the eiwres had a bug where it wasn't actually using the new distributions each time, so this also fixes the bug.
parallel _derived


ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/844

src/estimation/diagnostics.jl
```diff
@@ -333,15 +338,16 @@ function eiwres(m::PumasModel,
                 nsim::Integer,
                 args...;
                 kwargs...)
-  dist = _derived(m, subject, param, sample_randeffs(m, param), args...; kwargs...)
+
+  _population = [copy(subject) for i in 1:nsim]
+  dist = _derived(m, subject, param, args...; kwargs...)
```
andreasnoack:
```suggestion
  dist = _derived(m, _population, param, args...; kwargs...)

```

src/estimation/diagnostics.jl
```diff
@@ -333,15 +338,16 @@ function eiwres(m::PumasModel,
                 nsim::Integer,
                 args...;
                 kwargs...)
-  dist = _derived(m, subject, param, sample_randeffs(m, param), args...; kwargs...)
+
+  _population = [copy(subject) for i in 1:nsim]
```
andreasnoack: Why is the copy necessary here?

src/models/model_api.jl
```diff
@@ -208,6 +208,30 @@ to be repeated in the other API functions
   dist
 end
 
+@inline function _derived(m::PumasModel,
+                          pop::Population,
+                          param::NamedTuple,
+                          randeffs=sample_randeffs(m, param),
```
andreasnoack: I think this would need to be a vector of random effects for each subject in the population.

src/estimation/diagnostics.jl
ChrisRackauckas: I'm not sure it is, but I was going to first be safe and then investigate it.

src/models/model_api.jl
ChrisRackauckas: Yes like the simobs version. That's a bug.

ChrisRackauckas:
>I'm slightly concerned about the overhead. I tried to run some benchmarks and the overhead is noticeable. For npde with nsim=10000, I see about a single-threaded 200% slowdown for the proportional model in nlme/bolus.jl when using the closed-form model. When using four threads, the slowdown is 50%.

Where is the overhead? In the copying?

andreasnoack:
> Where is the overhead? In the copying?

I'm not sure where it is coming from. I suspect that some of it is the copying of the subject but maybe there is also something in setting up the ensemble problem.

Issue #817, opened 2020-01-17
Reproducibility of simobs


andreasnoack:
Currently, `simobs` isn't producing reproducible results by default when multithreading is enabled (which it is by default in JuliaPro). I.e. regardless of setting `Random.seed!` the results are not reproducible unless multithreading in `simobs` is disabled. Using the `simple_model` test model
```julia
julia> Random.seed!(1);

julia> map(subject -> subject.observed.dv[end], simobs(mdsl1, data, param, (η=[0.0],)))[end]
5.218673065123799

julia> Random.seed!(1);

julia> map(subject -> subject.observed.dv[end], simobs(mdsl1, data, param, (η=[0.0],)))[end]
5.451865151126203
```

but if multithreading is disabled

```julia
julia> map(subject -> subject.observed.dv[end], simobs(mdsl1, data, param, (η=[0.0],), ensemblealg=EnsembleSerial()))[end]
4.6121525241084385

julia> Random.seed!(1);

julia> map(subject -> subject.observed.dv[end], simobs(mdsl1, data, param, (η=[0.0],), ensemblealg=EnsembleSerial()))[end]
4.6121525241084385
```

This can be very confusing and many users will most likely expect reproducibility as the default.
```

ChrisRackauckas:
We can internally use local RNGs seeded by the id of the subject. Then if someone wants full randomness, we can have a kwarg flip that off. That would make it safe for all forms of parallelism.

PR #846, opened 2020-02-11
simobs multithreaded reproducibility by default

ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/817 by making the seed default to being based on the hash of the ID. This makes everything reproducible when parallel. This default can be disabled by `obsseed = nothing`, in which case the global rng takes hold and randomness ensues. This does have the unfortunate drawback that randomness is tied to the ID, so if all subjects have the same ID they will have the same randomness. I don't think this is too weird though, since ID is one of the required arguments in a Subject so it shouldn't show up... but it does show up in the diagnostics since we just create nsim copies of the same subject, and so in that case it's set to `obsseed=nothing` for full randomness.

And for timing:

```julia
julia> @btime rng = MersenneTwister(1234)
  10.500 μs (9 allocations: 19.48 KiB)
MersenneTwister(UInt32[0x000004d2], Random.DSFMT.DSFMT_state(Int32[-1393240018, 1073611148, 45497681, 1072875908, 436273599, 1073674613, -2043716458, 1073445557, -254908435, 1072827086  …  -599655111, 1073144102, 367655457, 1072985259, -1278750689, 1018350124, -597141475, 249849711, 382, 0]), [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0  …  0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0], UInt128[0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000  …  0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000, 0x00000000000000000000000000000000], 1002, 0)
```

that shouldn't be impactful.


andreasnoack:
From reading the diff, it wasn't clear to me that this approach would handle the draw of the random effects. Is that the case?

Even if it does, I'm not completely sold of this approach. Ideally, you should still be able to set the seed. As you say, this also affects the diagnostics and would exclude the reproducibility of simulated diagnostics by default.

My preference is still to disable parallelism by default and then document how to enable it with a warning note that it will make the results non-reproducible. Eventually, we might then be able to improve on that with something like http://supertech.csail.mit.edu/papers/dprng.pdf but that is a general Julia issue.

ChrisRackauckas:
>Even if it does, I'm not completely sold of this approach. Ideally, you should still be able to set the seed.

You can with this approach.

andreasnoack:
> You can with this approach.

I mean, change the seed to another reproducible seed. That won't be possible, right?

ChrisRackauckas:
`obsseed = 1:length(pop)`, `obsseed = length(pop)+1:2length(pop)`, etc.

vjd:
I recall from offline discussions that there was potentially a fix for this in Julia Base?

andreasnoack:
Might be https://github.com/PumasAI/Pumas.jl/issues/817#issuecomment-592179235. The linked PR is still WIP

Issue #750, opened 2019-11-22
Improve error message calling simobs without events


andreasnoack:
```julia
julia> mdsl1 = @model begin
           @param begin
               θ ∈ VectorDomain(1, init=[0.5])
               Ω ∈ RealDomain(lower=0.001, init=2.0)
           end
           @random begin
               η ~ Normal(0.0, Ω)
           end
           @pre begin
               CL = θ[1] * exp(η)
               V  = 1.0
           end
           @vars begin
               conc = Central / V
           end
           @dynamics ImmediateAbsorptionModel
           @derived begin
               dv ~ @. Normal(conc,0.11)
           end
       end
PumasModel
  Parameters: θ, Ω
  Random effects: η
  Covariates:
  Dynamical variables: Central
  Derived: conc, dv
  Observed: conc, dv


julia> simobs(mdsl1, Subject(obs=(dv=[1.0],), time=[0.0]), (init_param(mdsl1)))
ERROR: ArgumentError: reducing over an empty collection is not allowed
Stacktrace:
 [1] _empty_reduce_error() at ./reduce.jl:212
 [2] mapreduce_empty(::Function, ::Function, ::Type) at ./reduce.jl:246
 [3] _mapreduce(::Pumas.var"##538#539", ::typeof(promote_type), ::IndexLinear, ::Array{Pumas.Event,1}) at ./reduce.jl:301
 [4] _mapreduce_dim at ./reducedim.jl:312 [inlined]
 [5] #mapreduce#588 at ./reducedim.jl:307 [inlined]
 [6] mapreduce at ./reducedim.jl:307 [inlined]
 [7] sorted_approx_unique(::Array{Pumas.Event,1}) at /Users/andreasnoack/.julia/packages/Pumas/Ztp5C/src/simulate_methods/utils.jl:81
[...]

Issue #743, opened 2019-11-20
Mixed analytical and numerical model


pkofod:
It would be nice to have a way to specify a model that is part analytical, part numerical. One syntax suggestion is
```
@dynamics OneCmtModel begin
# use analytical compartment names in here
end
```

PR #847, opened 2020-02-11
Support MixedPK in the DSL

ChrisRackauckas:
Fixes https://github.com/PumasAI/Pumas.jl/issues/743 . 

Simultaneously fixes https://github.com/PumasAI/Pumas.jl/issues/750 since that makes it much easier to implement, as the PD portion doesn't have dosing events.


src/models/model_api.jl
```diff
@@ -133,10 +133,18 @@ function _problem(m::PumasModel, subject, col, args...;
     function _col(t)
       col_t = col(t)
       ___pk = pksol(t)
-      (col=col, ___pk=___pk)
+      (col_t..., Depot = ___pk.Depot, Central = ___pk.Central)
```
ChrisRackauckas: Needs an upstream fix for iterating a LabelledArray

ChrisRackauckas:
Just needs https://github.com/PumasAI/Pumas.jl/pull/847

src/dsl/model_macro.jl
```diff
@@ -630,8 +651,16 @@ macro model(expr)
     elseif ex.args[1] == Symbol("@init")
       extract_defs!(vars,ode_init, add_vars(ex.args[3], bvars))
     elseif ex.args[1] == Symbol("@dynamics")
-      isstatic = extract_dynamics!(vars, odevars, prevars, callvars, ode_init, ex.args[3], eqs)
-      odeexpr = ex.args[3]
+      if length(ex.args) == 3
```
pkofod: For clarity, could you mention what `args[3]` and `args[4]` is here? Just if someone else has to understand and edit this at some point, who is not you :)

ChrisRackauckas: yup I'll add a comment. IR is always hard, since it just puts it all in an unnamed `args` 🤷‍♂ . `@macro symbol begin ... end` will make `args[3]` be `symbol` and `args[4]` be `begin ... end`.

vjd:
@ChrisRackauckas bump!

PR #852, opened 2020-02-17
subject from simulation keeps observed quantities

ChrisRackauckas:
It is really weird that it currently drops the observation series in the generated subject. The natural thing to do for a simulated dataset would be to have the simulated observations be the simulated observations.

### DIFF
--- a/src/models/simulated_observations.jl
+++ b/src/models/simulated_observations.jl
@@ -14,12 +14,11 @@ end
 
 # Convert to Subject
 function Subject(simsubject::SimulatedObservations)
-  dvnames = keys(simsubject.subject.observations)
   covariates = simsubject.subject.covariates
   covartime = simsubject.subject.covartime
   subject = Subject(;
     id         = simsubject.subject.id,
-    obs        = NamedTuple{dvnames}(map(k -> simsubject.observed[k], dvnames)),
+    obs        = simsubject.observed,
     cvs        = covariates===nothing ? nothing : map(copy, covariates),
     evs        = copy(simsubject.subject.events),
     time       = copy(simsubject.times),
--- a/test/core/single_dosage_tests.jl
+++ b/test/core/single_dosage_tests.jl
@@ -68,7 +68,7 @@ m_analytic = @model begin
         V  = θ[3] * exp(η[2])
     end
 
-    @dynamics Depots1Central1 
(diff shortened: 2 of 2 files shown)


andreasnoack:
I don't think we should do this unless we require a match of the names in the `Subject`'s `observations` and the names in the `derived` block. Right now, we only use the intersection of the two name sets.

With this PR, we'd get weird roundtrip behavior, e.g. if you happen to write a model with a `derived` block like
```julia
@derived begin
      cp = @. Central / Vc
      dv ~ @. Normal(cp, cp*σ_prop)
end
```
and fit it on `Subject`s that just have `dv` in `observations, only `dv` would be considered in the likelihood calculations.

However, if you simulate from the model and convert the simulated subject to `Subject`s then the new subject would have multiple dvs. Therefore, you'd get completely different behavior when fitting the original subject and the new subject with the exact same model.

If we require a match between the names in the model and the subjects then the second fit would error instead of computing something very different from the original fit.

ChrisRackauckas:
Requiring a match wouldn't work well because different subjects can have different observation series. Fixing https://github.com/PumasAI/Pumas.jl/issues/396 would at least make the roundtrip work well, but what's really going on here is that the user wrote the wrong model. It sounds like they wanted:

```julia
@derived begin
      cp := @. Central / Vc
      dv ~ @. Normal(cp, cp*σ_prop)
end
```

but wrote the wrong model, and now we're trying to be too fancy and not do what they asked for hoping we are more correct than what they wrote down. That kind of feature always comes back to bite you, since then when they actually want to save `cp` and correctly do `=`, they can't do it because ... 🤷‍♂ magic in the background discarded it. I think we should just do what the user asks for and just make sure it's all sensible and explicit (here, `:=` has always meant "don't save", so not saving it to the observation series makes sense).

vjd:
Sorry for the delay in catching up on this. What seems to the be the exact issue here? 

As far as I understand, a model can have multiple outputs as part of the derived block and the user should be able to pick the right output to roundtrip. I tend to agree with @ChrisRackauckas here from whatever I understood the problem to be. (the part that I am not sure yet is what is @andreasnoack matching the output series to?, the variable in the derived block which has a `~` which naturally becomes the `observed` of a `Subject`). The user should have the flexibility to use the output of the derived variables as they want imho.

PR #854, opened 2020-02-19
Introduce option for threaded maximum likelihood estimation

andreasnoack:
The computation of the empirical Bayes estimates as well
as the computation of the gradient in the population parameters are
parallelized over Subjects.


pkofod:
So, it seems to work. There seems to be a limit to the speed-ups we can get, and they seme to be driven by allocations. We will have to investigate if we can improve this going forward.

PR #857, opened 2020-02-20
Avoid overwriting show methods for DosageRegimen

andreasnoack:
There were duplicates in the list of MIME types because the list of methods contained definitions from Base that applied to DataFrames objects.


andreasnoack:
Oops. That method of methods was only added in Julia 1.4.

ChrisRackauckas:
It looks like you're using `methods` and it works on 1.3?

andreasnoack:
In the original version of the PR, I was using the `methods(f, types, module)` methods but realized that the `module` argument was only added in Julia 1.4.

PR #872, opened 2020-02-25
Use a vector of randeffs if a population is given to simobs.

### LINKED ISSUE #867: simobs needs a method that accepts empirical_bayes(...) in long format
Currently you can simulate a single subject and specify their ebes, but you cannot do this for a complete population. Might also want a version defined on `FittedPumasModel` that accepts a keyword for whether to draw etas or use ebes, and maybe also to simulate from the fitted pumas model but under an alternative dosage regimen. Should put an Ensemble interface on this.

pkofod:
fixes #867 

Allow to `simobs` using the output of `empirical_bayes` (actually, I should add a test for that).

@ChrisRackauckas I just wanted to make sure that https://github.com/PumasAI/Pumas.jl/compare/simobs_ebes?expand=1#diff-04a2ed63ba1256374387642ee329b396R266 is correct. I mean, I maintained the existing functionality. Is it on purpose that if randeffs is `nothing` then the first function samples an eta while the second one passes on the `nothing` ? Is it not supposed to generate one eta and use it in both?


pkofod:
Okay, I see in #846 that they're not supposed to be the same. Then I guess this is the correct implementation.

src/estimation/likelihoods.jl
```diff
@@ -1021,6 +1021,7 @@ struct FittedPumasModel{T1<:PumasModel,T2<:Population,T3,T4<:LikelihoodApproxima
   kwargs::T7
   fixedtrf::T8
 end
+simobs(fpm::FittedPumasModel; kwargs...) = simobs(fpm.model, fpm.data, coef(fpm), empirical_bayes(fpm); kwargs...)
```
andreasnoack:
```suggestion
simobs(fpm::FittedPumasModel) = simobs(fpm.model, fpm.data, coef(fpm), empirical_bayes(fpm); fpm.kwargs...)
```

src/models/model_api.jl
```diff
@@ -263,7 +263,8 @@ function simobs(m::PumasModel, pop::Population,
     col = sol.prob.p
     obstimes = :obstimes ∈ keys(kwargs) ? kwargs[:obstimes] : observationtimes(pop[i])
     saveat = :saveat ∈ keys(kwargs) ? kwargs[:saveat] : obstimes
-    derived = m.derived(col,sol,obstimes,pop[i],param,randeffs)
+    _randeffs = randeffs === nothing ? nothing : randeffs[i]
+    derived = m.derived(col,sol,obstimes,pop[i],param,_randeffs)
```
andreasnoack:
```suggestion
    derived = m.derived(col, sol, obstimes, pop[i], param, _randeffs)
```

PR #873, opened 2020-02-26
Include dv's when called DataFrame on the output of predict

pkofod:
This is super useful for plotting the diagnostics.
fixes #873


src/estimation/diagnostics.jl
```diff
@@ -616,8 +617,9 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
       end
     end
   end
-  ebes_df[!, :ebes_approx] .= summary(i.ebes.approx)
+  ebes_df[!, :ebes_approx] .= fill(summary(i.ebes.approx), size(ebes_df, 1))
```
andreasnoack: Why was it necessary to add the `fill` here?

src/estimation/diagnostics.jl
```diff
@@ -616,8 +617,9 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
       end
     end
   end
-  ebes_df[!, :ebes_approx] .= summary(i.ebes.approx)
+  ebes_df[!, :ebes_approx] .= fill(summary(i.ebes.approx), size(ebes_df, 1))
   ebes_df = select!(select!(ebes_df, Not(:id)), Not(:time))
+  size(pred_df)
```
andreasnoack: This one shouldn't be there, right?

pkofod: nope :)

src/estimation/diagnostics.jl
pkofod: It tried to add the first letter to the first line, the second letter to the second line,... Maybe a `Ref` would do it?

PR #875, opened 2020-02-27
Temporarily add Compat@2 as a dependency to avoid tests failures

andreasnoack:
until all dependencies have been made compatible with Compat@3.

In a couple of tests, we are using `@test_nowarn` so when the tested code hits some of the deprecations in Compat@3 then they fail. This PR temporarily adds Compat@2 as a dependency to avoid the test failures. Once all dependencies are free from deprecation warnings, we can remove Compat again.


andreasnoack:
Our dependencies have been fixed so master now passes tests again.

PR #876, opened 2020-02-28
Enable LaplaceI for multiple dvs by deleting the exception and

andreasnoack:
reducing the chunk size used in the AD Hessian calculation.

Patrick will add tests.


test/nlme/runtests.jl
```diff
@@ -38,6 +38,7 @@ end
 if group == "All" || group == "NLME_ML3"
   @time @safetestset "Maximum-likelihood models 3" begin
     @time @safetestset "Time-varying covariates"                     begin include("tvcov.jl")                  end
+    @time @safetestset "Time-varying covariates"                     begin include("mdv.jl")                  end
```
pkofod:
```suggestion
    @time @safetestset "Multiple dependent variables"             begin include("mdv.jl")                  end
```

PR #879, opened 2020-03-06
Introduce coeftable to get a DataFrame of estimates

andreasnoack:
This PR introduces a `coeftable` function (extending StatsFuns) that returns a DataFrame with the estimation results. This DataFrame is then used by the show methods. With this, you can do
```julia
julia> coeftable(infer(fitFOCE))
Calculating: variance-covariance matrix. Done.
4×4 DataFrame
│ Row │ parameter │ estimate │ se        │ confint             │
│     │ String    │ Float64  │ Float64   │ Tuple…              │
├─────┼───────────┼──────────┼───────────┼─────────────────────┤
│ 1   │ θ₁        │ 3.44863  │ 0.837977  │ (1.80623, 5.09104)  │
│ 2   │ θ₂        │ 0.528904 │ 0.0792735 │ (0.37353, 0.684277) │
│ 3   │ ω         │ 1.03039  │ 0.139493  │ (0.756989, 1.30379) │
│ 4   │ θr        │ 0.945987 │ 0.0589291 │ (0.830488, 1.06149) │
```

As part of this PR, I've also changed relative standard errors to standard errors as discussed elsewhere since I touched that code anyway. Reporting standard errors in model output is the standard thing to do.


vjd:
this is really good. It will make it easy for us to write the notebooks and also for @asinghvi17 to compare the parameters across different models using a simple `join()`.

 Should we have a footnote as a table row at the end suggesting how the variance terms are being printed? i.e. are we printing the variance or the standard deviation and any special handling to compute the `se` for these variance terms?

andreasnoack:
> Should we have a footnote as a table row at the end suggesting how the variance terms are being printed?

It seems that there currently isn't proper support for metadata in DataFrames. Maybe we could consider appending `se` and `confint` with something indicating the method used, so something like `se_asymptotic` and `se_bootrap`. Just thinking out loud.

pkofod:
I think the usual way is just to have it be in the variable name, right? 
`coef_bootstrap` vs `coef_vcov` or similar

pkofod:
Andreas and I discussed that it would be useful to split the confint into two columns.

pkofod:
Great! Very useful.

PR #880, opened 2020-03-08
Soft fail on inference errors.

pkofod:
Let's get a soft failure on infer, it's about time! :)

Happy to take suggestions on the text and if it should only be PSD errors, but I just wanted to get the ball rolling. Not sure if we have an issue for this, but there are several confused discourse users!


vjd:
About time 😀

Simple and informative error. Obviously there may be other reasons, but we can grow them later

pkofod:
> Simple and informative error. Obviously there may be other reasons, but we can grow them later

Yes there may be other reasons. Andreas and I discussed it friday, and neither of us could remember a situation where it was *not*  a variance parameter for an eta that was the reason, and if that goes to zero, overparameterization is an obvious candidate - but other things might be wrong of course.

vjd:
LGTM

pkofod:
> First of all, we should definitely, have a better (error?) message when the `vcov` computation fails. The main question is if we should throw an exception or not. Implementing our own exception handling might become a problem but if we throw an exception then it forces users to wrap their `infer` calls in `try/catch` blocks.

I see what you're saying with the re-implementation. The idea was to be less scary, but we can throw an error instead I guess.

> Also, this will conflict with #879 so I'd appreciate if we could get that one merged first.

👍

pkofod:
I can't really make up my mind here.. Should it be an empty DataFrame? A DataFrame with NaN-entries?

pkofod:
Should work now.

PR #884, opened 2020-03-16
Fix empirical_bayes_dist and delete unused methods

andreasnoack:
This started out as a PR to increase the code coverage in the `estimation` folder. It revealed an issue with empirical_bayes_dist when the random effect covariance structure isn't diagonal. There were also some unused methods that I've deleted as well as a few other redundancies.


src/estimation/diagnostics.jl
```diff
@@ -552,26 +552,28 @@ function DataFrames.DataFrame(vpred::Vector{<:SubjectPrediction}; include_covari
...
 
   trf = totransform(fpm.model.random(coef(fpm)))
 
-  if approx == fpm.approx
+  if fpm.approx ∈ (Pumas.FOCE(), Pumas.FOCEI(), Pumas.LaplaceI())
     ebes = fpm.vvrandeffsorth
     return [TransformVariables.transform(trf, e) for (e, s) in zip(ebes, subjects)]
-  else
-    # re-estimate under approx
+  elseif fpm.approx === FO()
```
pkofod: Is this motivated by something specific? I'm not saying it's the wrong thing to do, just if it's what people will expect.

src/estimation/diagnostics.jl
```diff
@@ -616,7 +618,6 @@ function DataFrames.DataFrame(i::FittedPumasModelInspection; include_covariates=
       end
     end
   end
-  ebes_df[!, :ebes_approx] .= fill(summary(i.ebes.approx), size(ebes_df, 1))
```
pkofod: So this information is no loner in the daataframe, correct?

src/estimation/show.jl
```diff
@@ -18,18 +18,24 @@ function _print_fit_header(io, fpm)
   println(io)
 end
 
+
+function StatsBase.coeftable(nt::NamedTuple)
```
pkofod: :skull_and_crossbones: 

Is this kosher ? :) I agree it's a useful method.

src/estimation/diagnostics.jl
andreasnoack: The EBEs no longer depend on the approximation method so this option is redundant. We used to allow users to specify a model where the variance parameter in a Gaussian error model depended on the random effects but then ignore this dependence during estimation. Because of that, the EBEs sed to depend on the approximation method.

src/estimation/diagnostics.jl
andreasnoack: Correct

src/estimation/show.jl
andreasnoack: It's not so let me change this method to an internal `_coeftable` function.

src/estimation/diagnostics.jl
pkofod: Okay, so I was specifically thinking about the if FO then LaplaceI part.

andreasnoack: So for `FO` we evaluate the random effects at zero but that is still not the EBEs. That is just how the approximation works. The EBEs are still the maximizer of the penalized likelihood for fixed population parameters but this quantity isn't computed as part of the `FO` fitting. Therefore, it's done here by changing the approximation type to `LaplaceI`.

Issue #878, opened 2020-03-05
DataFrame output for NCAReport


vjd:
Currently, it is a two-step process to get the NCAReport as a DataFrame. 

1. `report = NCAReport(ncadata)`
2. `ncareport = NCA.to_dataframe(report)`

We should be able to consolidate this and just have NCAReport output a DataFrame such that the step below gives us a DataFrame

1. `report = NCAReport(ncadata)`

pkofod:
So is the Julia-representation never needed?

vjd:
not unless we are debugging anything. For a normal end-user, no

PR #885, opened 2020-03-17
Change `NCASubject` to a function that returns a DataFrame

YingboMa:
Fix https://github.com/PumasAI/Pumas.jl/issues/878


andreasnoack:
Please use a descriptive title and explain how the fix works. It makes it much easier to review the PR.

It looks like this makes the `NCAReport` constructor return a `DataFrame` which I don't think is a good idea. Is it still necessary to have an `NCAReport` struct or could `NCAreport`/`NCAReport` just be a function instead of a constructor.

pkofod:
Maybe keep the type (if it's useful and plays a role), don't pipe it into `DataFrame`, but instead have a shallow `ncareport(...) = DataFrame(NCAReport)` function?

pkofod:
LGTM

andreasnoack:
Same here. Any thoughts on the Markdown code?

vjd:
I think the DataFrame will render pretty well to markdown, so we can move that part out

src/nca/type.jl
```diff
@@ -426,24 +421,12 @@ function NCAReport(pop::NCAPopulation; pred=nothing, normalize=nothing, auctype=
...
-# units go under the header
-to_dataframe(report::NCAReport) = convert(DataFrame, report)
-DataFrames.DataFrame(report::NCAReport) = to_dataframe(report)
-function Base.convert(::Type{DataFrame}, report::NCAReport)
-  #report.settings[:subject] && return DataFrame(map(x->[x], report.values))
-  hcat(report.values...)
-end
+to_dataframe(report) = (@warn("`to_dataframe` is removed. `NCAReport` now returns a DataFrame directly."); report)
 
+#=
```
andreasnoack: Then let's just delete the Markdown code and this should be good to go.

PR #887, opened 2020-03-22
Fix normalized cmax

YingboMa:


andreasnoack:
It looks like the `interval` argument in `cmaxss` is also not passed.

YingboMa:
> It looks like the interval argument in cmaxss is also not passed.

That is intentional, steady-state `cmax` doesn't take intervals. @vjd could you confirm?

PR #888, opened 2020-03-23
Provide informative error message when a covariate contains only missing values.

andreasnoack:
The current error message, as reported in https://discourse.pumas.ai/t/reading-covariates-with-missing-data/306, isn't helpful.


andreasnoack:
I've added a commit that restricts CI to Julia 1.3 since the tests were failing because 1.4 has been released. We'll have to stay on Julia 1.3 for a bit until a compilation time regression in Julia 1.4 has been addressed.

pkofod:
I'm wondering: should we pass in the subject ID to build_tvcov so we can mention the ID in case it fails? Other than that looks ood.

vjd:
I agree with Patrick. We should provide the ID number too

